PR #3, opened 2020-02-14
Transpile ES6 with Babel.

liamzebedee:
Two fixes to get tbtc.js working within a dApp context:

 1. Transpiling the ES6 syntax. We're using a not-so-ubiquitous construct of class members (see below; as for why it's a special case, don't ask me) which needs some transpilation to work with eg. react-scripts

```
| class TBTC {
>     config/*: TBTCConfig*/;
|
|     constructor(config/*: TBTCConfig*/ = defaultConfig, networkMatchCheck = true) {
```

 2. Fix a line where we do a readFile.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -37,7 +37,7 @@ const VendingMachineContract = TruffleContract(VendingMachineJSON)
 const ECDSAKeepContract = TruffleContract(ECDSAKeepJSON)
 
 // TODO Need this configured via TBTC.
-const electrumConfig = JSON.parse(fs.readFileSync("./src/electrum-config.json"))
+const electrumConfig = require(__dirname + "/../src/electrum-config.json")
 
 export class DepositFactory {
     config/*: TBTCConfig*/;
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+    "presets": [
+        "@babel/preset-env"
+    ],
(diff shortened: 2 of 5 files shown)


Shadowfiend:
Turns out this is experimental, whoops. Two things coming up:
- Dropping the config file in favor of passing in the Electrum config (that will eliminate the need for `readFileSync` completely).
- Leaving the field declarations purely as comments/maybe as JSDoc annotations.

Right now my goal for this lib is to avoid all Babelification and stick to vanilla JS; that may change depending on what we hear back from users, but let's see if we can hold that line.

src/Deposit.js
```diff
@@ -37,7 +37,7 @@ const VendingMachineContract = TruffleContract(VendingMachineJSON)
 const ECDSAKeepContract = TruffleContract(ECDSAKeepJSON)
 
 // TODO Need this configured via TBTC.
-const electrumConfig = JSON.parse(fs.readFileSync("./src/electrum-config.json"))
+const electrumConfig = require(__dirname + "/../src/electrum-config.json")
```
Shadowfiend: As a side note, this in turn doesn't work in node proper because `require` isn't allowed from ES6 modules, at least on older versions of node.

Delightful!

Shadowfiend:
Landed this on master for 0.6.5; closing the PR for now.

PR #6, opened 2020-02-25
Typing Day: Deep pass at JSDoc-ing types

Shadowfiend:
> [**Typing Day** (also known as **International Typing Day** or **World Typing Day**) is an annual event that falls on 8 January in Malaysia. It is co-organized by the STC (Speed Typing Contest) Team from JCI (Junior Chamber International) Mines and Team TAC (Typo Auto Corrector) to promote speed, accuracy and efficiency in written communication among the public.](https://en.wikipedia.org/wiki/Typing_Day)

This goes through a large swath of the functionality of the library and
annotates types throughout using JSDoc. Editors like VSCode and
compilers like TypeScript can pull this information, implicitly or
sometimes explicitly through /** import */ declarations, alongside
`@ts-check` annotations, to typecheck code that uses tbtc.js, either in
TypeScript or in regular JS.

-----------

The goal here isn't to be 100% complete, but to get close and hit most of the common code paths.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -66,7 +66,7 @@ export class DepositFactory {
      * @return {Promise<BN[]>} A list of the available lot sizes, in satoshis,
      *         as BN instances.
      */
-    async availableSatoshiLotSizes()/*: Promise<BN[]>*/ {
+    async availableSatoshiLotSizes() {
         return await this.systemContract.getAllowedLotSizes()
     }
 
@@ -83,7 +83,7 @@ export class DepositFactory {
      * 
      * @return {Promise<Deposit>} The new deposit with the given lot size.
      */
-    async withSatoshiLotSize(satoshiLotSize/*: BN*/)/*: Promise<Deposit>*/ {
+    async withSatoshiLotSize(satoshiLotSize) {
         if (! await this.systemContract.isAllowedLotSize(satoshiLotSize)) {
             throw new Error(
                 `Lot size ${satoshiLotSize} is not permitted; only ` +
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -1,8 +1,10 @@
 import BitcoinHelpers from "./BitcoinHelpers.js"
+/** @typedef { import("./BitcoinHelpers.js").FoundTransaction } BitcoinTransaction */
 
 import EthereumHelpers from "./EthereumHelpers.js"
 
-import TruffleContract from "@truffle/contract"
+import TruffleContract from '@truffle/contract'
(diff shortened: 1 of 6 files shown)


src/Deposit.js
```diff
@@ -210,16 +293,26 @@ export default class Deposit {
         return new Deposit(factory, contract, keepContract)
     }
 
-    static async forTDT(factory/*: DepositFactory*/, tdt/*: TBTCDepositToken | string*/)/*: Promise<Deposit>*/ {
+    /**
+     * @param {DepositFactory} factory 
+     * @param {any | string} tdt
```
mhluongo: Is it useful to have a sum type with `any` as a constituent? Forgive me if I'm misreading this

src/TBTC.js
```diff
@@ -1,43 +1,53 @@
 import { DepositFactory } from "./Deposit.js";
 
+/** @enum {string} */
```
mhluongo: This is nice

mhluongo:
@liamzebedee can you give this another and mere if it's good?

src/Deposit.js
Shadowfiend: This function is still unimplemented, but don't tell the project lead :wink:

Shadowfiend: (If you look at the original type annotation, my intent is for this to be able to take an instance of `TBTCDepositToken`, but that's a wrapper class that doesn't exist yet.)

mhluongo: Shwew, just checking my sanity 👌

Shadowfiend: Yeah no, great callout. `any | string` is about as useful as an empty pizza box.

liamzebedee:
@Shadowfiend looks like I've got no write access. 😬

src/TBTC.js
```diff
@@ -1,43 +1,53 @@
...
-/*
-export type TBTCConfig = {
-    bitcoinNetwork: BitcoinNetwork,
-    web3: Web3,
-}
-*/
+/**
+ * @typedef {Object} TBTCConfig
+ * @prop {BitcoinNetwork} bitcoinNetwork
+ * @prop {Web3} web3
```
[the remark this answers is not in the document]
Shadowfiend: Web3 isn't a direct dependency right now; when you move us to Web3 contracts that should do the trick.

mhluongo:
I'll merge when these boys are resolved

Shadowfiend:
Everyone on Keep now has write access, pushed fixes for the above bits.

PR #7, opened 2020-02-25
Iron Confederacy: Improve tbtc.js for usage by the tBTC reference dApp

Shadowfiend:
> [Poundmaker (1842 – 4 July 1886), born Pîhtokahanapiwiyin, was a Plains Cree chief. He leads the Cree in Civilization VI: Rise and Fall.](https://civilization.fandom.com/wiki/Poundmaker_(Civ6))
> …
> [Poundmaker's unique agenda is called **Iron Confederacy**. He tries to establish as many alliances as possible, and dislikes civilizations that don't establish alliances.](https://civilization.fandom.com/wiki/Poundmaker_(Civ6))

While porting the reference dApp from its haphazard internal library to tbtc.js, several missing hooks and API cleanups became apparent. This PR implements those hooks and cleans up those APIs, including:

- Access to a deposit's current state.
- Access to BTC-satoshi and TBTC-satoshi conversion factors.
- Tweaks in the behavior of `getCurrentRedemption`.
- Exposure to the intermediate steps of auto-submission for both deposit creation and redemption flows.

It's a smashing good time 👊

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -26,6 +26,34 @@ const FeeRebateTokenContract = TruffleContract(FeeRebateTokenJSON)
 const VendingMachineContract = TruffleContract(VendingMachineJSON)
 const ECDSAKeepContract = TruffleContract(ECDSAKeepJSON)
 
+/** @enum {number} */
+const DepositStates = {
+    // Not initialized.
+    START: 0,
+
+    // Funding flow.
+    AWAITING_SIGNER_SETUP: 1,
+    AWAITING_BTC_FUNDING_PROOF: 2,
+
+    // Failed setup.
+    FRAUD_AWAITING_BTC_FUNDING_PROOF: 3,
+    FAILED_SETUP: 4,
+
+    // Active/qualified, pre- or at-term.
+    ACTIVE: 5,
+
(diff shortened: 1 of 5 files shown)


src/Deposit.js
```diff
@@ -687,7 +733,12 @@ export default class Deposit {
 
     ///------------------------------- Helpers ---------------------------------
 
-    // autoSubmitting/*: boolean*/
+    /**
+     * @typedef {Object} AutoSubmitState
+     * @prop {Promise<BitcoinTransaction>} fundingTransaction
+     * @prop {Promise<{ transaction: FoundTransaction, requiredConfirmations: Number }>} fundingConfirmations
```
liamzebedee: Tiny observation - exposing `requiredConfirmations` *after* we've waited those confirmations feels a bit confusing at first.

Shadowfiend: Yeah… I file this under “necessary but annoying”, because the next step (proof) uses the required confirmations to limit the SPV proof to the required amount of work, rather than submitting all work on the transaction. That lets us avoid issues with submitting older transactions whose proof data ends up being too large.

PR #9, opened 2020-03-11
Lintify: Introduce some of that sweet sweet linting and prettying

Shadowfiend:
New new:
- CI that runs linting and tests.
- Linting that does format checking + lint fixing that formats as well.
- Pre-commit config.

------------

Still missing here:
- [x] CI check.
- [x] Merging + tagging keep-network/eslint-config-keep#4 and updating the reference here to a tagged version.


Shadowfiend:
Okayyy turbo-merging here since everything looks good. Yay CI!

PR #12, opened 2020-03-25
Tranche Deux: Add helper for converting ABI-coded `bytes` into their underlying bytes

liamzebedee:
This makes the code easier to read for Bitcoiners, and the like, without having to understand the specifics of the EVM ABI coding.

Follow-up from #10.

### DIFF
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -81,8 +81,19 @@ async function getExistingEvent(source, eventName, filter) {
   return events[0]
 }
 
+/**
+ * Converts an Ethereum `bytes` value into the raw bytes it represents.
+ * Drops the 0x prefix, and the length prefix.
+ * @param {string} bytesString An Ethereum-encoded `bytes` string
+ * @return {string} The hexadecimal string.
+ */
+function bytesToRaw(bytesString) {
+  return bytesString.replace("0x", "").slice(2)
+}
+
 export default {
   getEvent,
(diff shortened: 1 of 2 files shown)


src/EthereumHelpers.js
```diff
@@ -81,8 +81,19 @@ async function getExistingEvent(source, eventName, filter) {
   return events[0]
 }
 
+/**
+ * Converts an Ethereum `bytes` value into the raw bytes it represents.
+ * Drops the 0x prefix, and the length prefix.
+ * @param {string} bytesString An Ethereum-encoded `bytes` string
+ * @return {string} The hexadecimal string.
```
Shadowfiend: Had a quick look at `BitcoinHelpers`; I think to stay consistent, we want to say “The hexadecimal string without `0x` or length prefixes.”

My original suggestion on the function naming (`EthereumHelpers.spvToBitcoinOutputScript`) was deliberately specific because this is really rooted in the particular vagaries of how bcoin takes something on one end and how *we specifically represent that same information in Ethereum*.

That said, I struggled a little to articulate it in a method name, and I'm fine with living with this name for now and renaming later if we find it problematic, confusing, or frustrating.

liamzebedee: Oh thanks! I honestly wasn't even aware of the need for slicing the length prefix byte, so the name probably originates from that beginner's mindset ;)

PR #15, opened 2020-04-02
Fixes Shmixes: Fix a variety of details from recent merges

Shadowfiend:
Hitting a few corners in the dApp, several fixes became apparent.
- New state identifiers have landed in tbtc.
- getRedemptionCost wasn't always returning a `BN`.
- `IBondedECDSAKeep` didn't include the events needed for the deposit process.
- Transaction sends didn't have a way of resolving revert messages.

It turns out transaction sends still don't really have a way of resolving
revert messages, but there is now a wrapper in place that can be iterated
on to get those messages figured out.


src/Deposit.js
```diff
@@ -338,7 +330,7 @@ export default class Deposit {
     const keepAddress = createdEvent.returnValues._keepAddress
     console.debug(`Found keep address ${keepAddress}.`)
     const keepContract = new web3.eth.Contract(
-      IBondedECDSAKeepJSON.abi,
+      BondedECDSAKeepJSON.abi,
```
liamzebedee, line 333: 👍

src/Deposit.js
```diff
@@ -613,11 +604,13 @@ export default class Deposit {
 
       return lotSize.mul(toBN(10).pow(toBN(10))).add(ownerRedemptionRequirement)
     } else {
-      return await this.contract.methods
-        .getRedemptionTbtcRequirement(
-          this.factory.config.web3.eth.defaultAccount
-        )
-        .call()
+      return toBN(
```
liamzebedee, line 607: 👍

src/EthereumHelpers.js
```diff
@@ -91,9 +91,64 @@ function bytesToRaw(bytesString) {
...
+ * @param {*} sendParams The parameters to pass to `estimateGas` and `send` for
+ *        transaction processing.
+ * @param {boolean} forceSend Force the transaction send through even if gas
+ *        estimation fails.
+ *
+ * @return {Promise<any>} A promise to the result of sending the bound contract
+ *         method. Fails the promise if gas estimation fails, extracting an
+ *         on-chain error if possible.
+ */
+async function sendSafely(boundContractMethod, sendParams, forceSend) {
```
liamzebedee, line 113: `forceSend` feels like a workaround for an issue with gas estimation. I haven't encountered it yet, it would be good to document it a bit more here / in an issue.

liamzebedee: The error handling sounds like it could be solved by setting [`handleRevert`](https://web3js.readthedocs.io/en/v1.2.6/web3-eth-contract.html#handlerevert) in `initializeContracts`, which just came to light. However, happy to consider it nonblocking if there are higher priority things for now.

src/EthereumHelpers.js
```diff
@@ -91,9 +91,64 @@ function bytesToRaw(bytesString) {
...
+    const gasEstimate = await boundContractMethod.estimateGas(sendParams)
+
+    return boundContractMethod.send({
+      gas: gasEstimate,
+      ...sendParams
+    })
+  } catch (exception) {
+    // If we're not forcibly sending, try to resolve the true error by using
+    // `call`.
+    if (exception.message.match(/always failing transaction/)) {
```
liamzebedee, line 124: Should we rethrow `exception` if we don't enter this block?

src/Deposit.js
```diff
@@ -15,7 +15,7 @@ import DepositFactoryJSON from "@keep-network/tbtc/artifacts/DepositFactory.json
 import TBTCTokenJSON from "@keep-network/tbtc/artifacts/TBTCToken.json"
 import FeeRebateTokenJSON from "@keep-network/tbtc/artifacts/FeeRebateToken.json"
 import VendingMachineJSON from "@keep-network/tbtc/artifacts/VendingMachine.json"
-import IBondedECDSAKeepJSON from "@keep-network/tbtc/artifacts/IBondedECDSAKeep.json"
+import BondedECDSAKeepJSON from "@keep-network/keep-ecdsa/artifacts/BondedECDSAKeep.json"
```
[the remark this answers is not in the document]
Shadowfiend: I did this while doing the version tether work… Didn't get a chance to test it until I wired the dApp up, and realized the flub.

src/EthereumHelpers.js
Shadowfiend: Yes please, definitely let's not block, though I'll likely introduce a PR this morning to add that back.

Agree re: gas estimation. Still don't know what's up there, and perhaps `handleRevert` will help.

src/EthereumHelpers.js
Shadowfiend: Yep, great flag. Fixed.

PR #21, opened 2020-04-21
approveAndCall for funding

liamzebedee:
This introduces `approveAndCall` for funding, which facilitates proving and minting TBTC in one transaction. 

## Context
To mint TBTC from an unqualified Deposit, we need to make three Ethereum transactions:

 1. `Deposit.provideFundingProof`
 2. `TDTToken.approve(VendingMachine, tdtId)`
 3. `VendingMachine.tdtToTbtc`

We have implemented the `approveAndCall` pattern to simplify this down to one call. The approval transaction (2) is avoided. A wrapper method, 
`VendingMachine.unqualifiedDepositToTbtc`, performs calls (1) and (3). Due to implementation details of our system, we use a wrapper script, `RedemptionScript`, to call the VendingMachine.

## Design
Currently, users can use `autoSubmit` to automatically provide the funding proof, and then call `mintTBTC` to mint their TBTC. These entail the three separate transactions above.

@Shadowfiend and I discussed the API on Keybase. I proposed an API like `autoSubmit({ mintTBTC: boolean = true })`, with the default being to mint TBTC. We decided a separate method would be better, `autoMint`.

If the user wishes to keep their TDT, they can call `deposit.autoSubmit()`. If they wish to mint TBTC, they call `deposit.autoMint()`.

## Refactoring
`autoMint` relies on some shared events of `autoSubmit`: `fundingTransaction` and `fundingConfirmations`. 

These were extracted into class members, that can be shared and used by both our methods and external users alike. eg. third-party TDT buyers can `await deposit.fundingTransaction` to avoid dealing with the bitcoin network complexity.

Deps: https://github.com/keep-network/tbtc/pull/594 

Closes https://github.com/keep-network/tbtc.js/pull/21, https://github.com/keep-network/tbtc.js/issues/26


Shadowfiend:
> Linting

As a side note, did you install the precommit hook? We shouldn't have any linting commits as they should be caught before a commit lands.

liamzebedee:
> It seems like autoMint could have been implemented as return autoSubmit().then({ result => result.transaction ... }) or so

Doing this would entail three Ethereum transactions: one call to `provideFundingProof` in `autoSubmit`, and the second/third one to approve the TDT for transfer, and then calling `tdtToTbtc`. 

The rationale for using FundingScript/approveAndCall is to simplify this into one transaction. `approveAndCall` will approve the TDT token, and proving + minting TBTC is accomplished in one call in `VendingMachine.unqualifiedDepositToTbtc`. 

Sorry, I could've added more context on this to support your review. I'll add some above.

src/Deposit.js
```diff
@@ -811,52 +880,75 @@ export default class Deposit {
     }
     /** @type {AutoSubmitState} */
     const state = (this.autoSubmittingState = {})
+    state.fundingTransaction = this.fundingTransaction
+    state.fundingConfirmations = this.fundingConfirmations
```
Shadowfiend, line 884: I wonder if we should include these in the state at all given they're now exposed directly on the deposit… My original thinking I believe was that we not lock the deposit into a single minting/confirmation/etc call, thus loading it all into this state object, but I kind of prefer the setup in this PR.

Dropping it from the auto-submit state can wait, though, haven't thought through all the implications yet.

src/Deposit.js
```diff
@@ -811,52 +880,75 @@ export default class Deposit {
...
     )
 
-    state.proofTransaction = state.fundingConfirmations.then(
-      async ({ transaction, requiredConfirmations }) => {
+    return state
+  }
+
+  autoMint() {
+    // Only enable auto-submitting once.
+    if (this.autoMintingState) {
```
Shadowfiend, line 910: Should we unify this and `autoSubmitState`? Feels like we shouldn't allow auto-submitting and auto-minting in parallel.

src/Deposit.js
```diff
@@ -568,9 +626,20 @@ export default class Deposit {
...
     )
     proofArgs.unshift(this.address)
-    const transaction = await this.factory.vendingMachineContract.methods
-      .unqualifiedDepositToTbtc(...proofArgs)
-      .send()
+
+    // Use approveAndCall pattern to execute VendingMachine.unqualifiedDepositToTbtc.
+    const unqualifiedDepositToTbtcCall = this.factory.vendingMachineContract.methods
+      .unqualifiedDepositToTbtc(this.address, ...proofArgs)
+      .encodeABI()
```
Shadowfiend, line 633: Can't we drop the `approve` call above now?

src/Deposit.js
```diff
@@ -568,9 +626,20 @@ export default class Deposit {
...
+
+    const transaction = await EthereumHelpers.sendSafely(
+      this.factory.tokenContract.methods.approveAndCall(
```
Shadowfiend, line 636: Are we approving transfer on the token contract or the deposit token contract?

src/Deposit.js
liamzebedee: Happy to do this. Needs a better name I feel?

src/Deposit.js
liamzebedee: > My original thinking I believe was that we not lock the deposit into a single minting/confirmation/etc call, thus loading it all into this state object, but I kind of prefer the setup in this PR.

How about intro'ing a `fundingConfirmationsFor` with the transaction as a parameter eg.

```js
// manual usage
const transaction = await deposit.fundingTransaction
await deposit.fundingConfirmationsFor(transaction)

// in Deposit
get fundingConfirmations() {
    return this.fundingTransaction.then(transaction => this.fundingConfirmationsFor(transaction))```
}

src/Deposit.js
liamzebedee: Just swapped from using git stashes to 'wip' branches. This must've slipped through, woops.

liamzebedee:
@Shadowfiend addressed all of the above, and made the Git history much more seemly.

ironng:
@Shadowfiend I tested both the deposit and redemption flows and made one fix. This should be good to go now. Just note, we'll need to update the tbtc version dependency to the latest (1.1.0-pre?), since we made updates for the notification station prs. I guess we can do that later, once we make the tbtc release?

Shadowfiend:
Yeah, I think that's what'll happen.

PR #23, opened 2020-04-23
Unblock build and publish for redeploy

Shadowfiend:


Shadowfiend:
This is pretty wildly outdated, putting it to bed.

PR #24, opened 2020-04-24
Fix example.js

liamzebedee:
web3 requires a different provider setup than `@truffle/hdwallet-provider`. This PR pulls in `@0x/subproviders` and `web3-provider-engine` to get `bin/tbtc.js`, `example.js`, and `resume.js` running again.

### DIFF
--- a/resume.js
+++ b/resume.js
@@ -1,13 +1,20 @@
 import Web3 from "web3"
 import TBTC from "./src/TBTC.js"
 
-import HDWalletProvider from "@truffle/hdwallet-provider"
-const mnemonic =
-  "egg dune news grocery detail frog kiwi hidden tuna noble speak over"
+import ProviderEngine from "web3-provider-engine"
+import Subproviders from "@0x/subproviders"
 
-const provider = new HDWalletProvider(
-  mnemonic,
-  "https://:e18ef5ef295944928dd87411bc678f19@ropsten.infura.io/v3/59fb36a36fa4474b890c13dd30038be5"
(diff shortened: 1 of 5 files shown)


Shadowfiend:
You need to commit a `package-lock` update though.

liamzebedee:
Updated the other uses too. Realised today I'll have to update the provider again to be a websockets provider. Initially a bit frustrating, but I tested it and it's actually noticeably quicker than the HTTP transport.

resume.js
```diff
@@ -1,21 +1,29 @@
...
 )
 
 const depositAddress = "0xb35671830Ee9E6651D1486b32279FB144D449c94"
 // const depositAddress = "0x89A2FBB780BFEa3aC4bC4Fed2b6147dEeB66BD58"
 const redeemerAddress = "tb1qdajvg4waymq604gpfvjdvpxyd2hc4yd3u4lse5"
 
 async function runExample() {
-  const web3 = new Web3(provider)
+  const web3 = new Web3(engine)
+  engine.start()
```
Shadowfiend, line 26: What even, but okay.

liamzebedee:
@Shadowfiend I'd reaaaaaally like to move to ethers.js after the past fortnight. Only blocker right now, is it might be a bit of work to integrate providers (eg. Coinbase, Fortmatic, etc). Haven't checked it out in depth, but just putting it out there.

Shadowfiend:
Noted. Timeline’s tight, but definitely something to consider.

PR #25, opened 2020-04-27
Liquidation wrappers

liamzebedee:
This PR exposes the liquidation methods on tBTC Deposits. 

 - [x] notifySignerSetupFailure
 - [x] notifyFundingTimeout
 - [x] notifySignatureTimeout
 - [x] notifyRedemptionProofTimeout
 - [x] notifyUndercollateralizedLiquidation
 - [x] notifyCourtesyCall
 - [x] notifyCourtesyTimeout
 - [x] exitCourtesyCall
 - [x] provideECDSAFraudProof
 - [x] provideFundingECDSAFraudProof
 - [x] purchaseSignerBondsAtAuction

I've also developed a mini-maintainer agent. We can shift this into another PR if need be, it's mostly for testing and guiding the design.

To do:

 - [x] API for accessing TBTCConstants from tbtc.js
 - [x] Splitting up PR into helpers and maintainer
 - [x] Address ElectrumHelpers changes.


src/Deposit.js
```diff
@@ -50,6 +50,12 @@ const DepositStates = {
   LIQUIDATED: 11
 }
 
+function getDepositState(i) {
+  return Object.values(DepositStates)
+    .filter(([state, j]) => i == j)[0]
+    .toString()
```
Shadowfiend, line 56: I wonder if we just export `DepositStates` instead?

src/Deposit.js
```diff
@@ -1016,4 +1022,62 @@ export default class Deposit {
       digest: _digest
     }
   }
+
+  /**
+   * Get the current collateralization level for this Deposit
+   * @return {number} Percentage collateralisation, as an integer. eg. 120%
+   */
+  async getCollateralisation() {
```
mhluongo, line 1030: British spelling alert. Calling out because it's inconsistent

liamzebedee: Oops.

src/Deposit.js
```diff
@@ -438,6 +438,10 @@ export default class Deposit {
     return toBN(await this.contract.methods.lotSizeSatoshis().call())
   }
 
+  async getTBTCLotSize() {
+    return toBN(await this.contract.methods.lotSizeTbtc().call())
+  }
```
Shadowfiend, line 443: … This might be the right moment to rename these to match what we're using on the contract side. No idea why I didn't do this in the first place 🤔

src/Deposit.js
```diff
@@ -1099,4 +1103,174 @@ export default class Deposit {
...
+    await EthereumHelpers.sendSafely(
+      this.contract.methods.notifyUndercollateralizedLiquidation()
+    )
+  }
+
+  /**
+   * Purchases the signer bonds and closes the liquidation auction.
+   *
+   * Note: The caller must approve the Deposit contract to spend
+   * `lotSize` amount of TBTC.
```
Shadowfiend, line 1194: This feels like it should be built into this particular function. If people want to call the underlying contract function independently of the “blessed path”, they can and should call `.contract` and interact with its methods directly.

src/Deposit.js
```diff
@@ -1099,4 +1103,174 @@ export default class Deposit {
...
+    )
+  }
+
+  /**
+   * Gets the current value of signer bonds at auction.
+   * Only callable if the deposit is in the liqudation state.
+   * @return {BN} auction value in wei.
+   */
+  async auctionValue() {
+    return toBN(await this.contract.methods.auctionValue().call())
```
Shadowfiend, line 1208: We should probably either return numbers or BNs, but do so consistently (or BigNumbers, or whatever).

src/Deposit.js
```diff
@@ -1099,4 +1103,174 @@ export default class Deposit {
...
+      this.contract.methods.notifyRedemptionProofTimeout()
+    )
+  }
+
+  /**
+   * Checks if signature was requested via the Keep.
+   * @param {string} digest Digest to check approval for.
+   * @return {boolean} True if signature approved, false if not (fraud).
+   */
+  async wasSignatureApproved(digest) {
```
Shadowfiend, line 1267: I think we can cheat this one by calling `isAwaitingSignature(digest)` on the backing keep.

Shadowfiend: Though I suppose this will false-negative if this signature was approved, but previously (e.g., in case of fee bump). So 50-50 on that.

src/Deposit.js
```diff
@@ -1099,4 +1103,174 @@ export default class Deposit {
...
+    return toBN(await this.contract.methods.auctionValue().call())
+  }
+
+  // /--------------------- Timeout helpers -------------------------
+
+  /**
+   * Notify the contract that signing group setup has timed out.
+   * Only applicable during funding.
+   */
+  async notifySignerSetupFailure() {
```
Shadowfiend, line 1217: Definitely what's driving me bananas about all these helpers is that we're basically reimplementing what `TruffleContract` gave us for free :/

src/Deposit.js
liamzebedee: Mmmmm, is there any reason against using number solely for the collateralisation thresholds? It reads sooo much more cleanly.

src/Deposit.js
liamzebedee: Yeaaah, that'd be nice.

src/Deposit.js
liamzebedee: Yep yep yep 💯 Absolutely feel this horror of verbosity.

I could've done with conducting a stronger investigation into web3 beforehand, maybe testing the waters rather than diving right in. ethers.js seems a lot easier to work with in retrospect.

src/Deposit.js
Shadowfiend: Hmm… How so? My main concern is that it makes our own return values mutually incompatible. So if `getTBTCLotSize` returns a `BN` and you want to apply a `number` collateralization threshold to it (say, for display purposes), you have to convert the `number` to a `BN`. Ditto for `getSatoshiLotSize`. It's more for folks to keep mental track of---am I interacting with a `BN` or a `number`? While I don't love the `BN` API (+ limitations), I like having to track disparate types even less.

And while `collateralisation < thresholdCourtesy` is more readable than `collateralisation.lt(thresholdCourtesy)`, I feel like it's not enough so to make the tradeoff worth it.

src/Deposit.js
Shadowfiend: Aye, though it's also worth remembering that the whole point of our library is giving a consistent interface to our contracts, so it's natural to have to untangle some of the underlying impedance mismatches ourselves.

mhluongo: > ethers.js seems a lot easier to work with in retrospect.

Can confirm that ethers.js is easier for the happy path, but when you need to do something different- eg an unsigned transaction for offline signing- the interface breaks down a bit

src/Deposit.js
liamzebedee: > So if getTBTCLotSize returns a BN and you want to apply a number collateralization threshold to it (say, for display purposes), you have to convert the number to a BN.

Ah, of course. Yeah, BN-to-go it is.

src/Deposit.js
liamzebedee: Yeah I don't believe the method is giving us what we'd want - the aim is to use wasSignatureApproved to check fraud. Checking if it's approved would be strictly different.

Shadowfiend:
Let's do split out the mini-maintainer into its own PR (or perhaps 2?) in whatever way makes the most sense for you. This is going to be a beast to review, fed thoroughly by how long it took me to get around to it, but I want to get to each piece in turn.

liamzebedee:
@Shadowfiend Liquidation helpers have been extracted out from the mini-maintainer. Should be legible for review now.

ironng:
@Shadowfiend Tested both deposit and redemption flows locally. Everything works :)

PR #34, opened 2020-05-21
Added web3-utils to dependencies

pdyraga:
When trying to use tbtc.js with dApp, the compilation failed with:

```
../tbtc.js/src/Deposit.js
Module not found: Can't resolve 'web3-utils' in '/Users/piotr/git/local-setup/tbtc.js/src'
```

All `web3*` dependencies are now placed in `peerDependencies`.


package.json
```diff
@@ -28,12 +28,13 @@
...
   "dependencies": {
-    "@keep-network/tbtc": ">0.15.0-pre <0.15.0-rc",
     "@keep-network/keep-ecdsa": ">0.15.0-pre <0.15.0-rc",
+    "@keep-network/tbtc": ">0.15.0-pre <0.15.0-rc",
     "bcoin": "git+https://github.com/bcoin-org/bcoin.git#8851582",
     "bcrypto": "^4.1.0",
     "bufio": "^1.0.6",
-    "electrum-client-js": "git+https://github.com/keep-network/electrum-client-js.git#v0.1.0"
+    "electrum-client-js": "git+https://github.com/keep-network/electrum-client-js.git#v0.1.0",
+    "web3-utils": "^1.2.8"
```
Shadowfiend, line 37: I think we actually need to move web3 and its provider packages to `peerDependencies`. The binaries that come with tbtc.js require all of those to work as well.

PR #36, opened 2020-05-28
Support the btcd "simnet"

liamzebedee:
simnet is the equivalent of a regtest network in bitcoind. This adds changes such that we encode/display addresses that can be used with btcd's simnet mode.

### DIFF
--- a/src/TBTC.js
+++ b/src/TBTC.js
@@ -52,6 +52,8 @@ export class TBTC {
       networkMatchCheck &&
       ((isMainnet(config.web3) &&
         config.bitcoinNetwork == BitcoinHelpers.Network.TESTNET) ||
+        (isMainnet(config.web3) &&
+          config.bitcoinNetwork == BitcoinHelpers.Network.SIMNET) ||
         (isTestnet(config.web3) &&
           config.bitcoinNetwork == BitcoinHelpers.Network.MAINNET))
     ) {
--- a/src/TBTC.js
+++ b/src/TBTC.js
@@ -1,6 +1,7 @@
 import { DepositFactory } from "./Deposit.js"
 import BitcoinHelpers from "./BitcoinHelpers.js"
 import BN from "bn.js"
+import bcoin from "bcoin"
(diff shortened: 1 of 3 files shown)


src/TBTC.js
```diff
@@ -69,6 +70,9 @@ export class TBTC {
     this.config = config
 
     this.satoshisPerTbtc = new BN(10).pow(new BN(10))
+    // Set default bcoin network.
+    // This affects how addresses are encoded and displayed.
+    bcoin.set(config.bitcoinNetwork)
```
Shadowfiend, line 75: I think instead of this we need to extend support in `BitcoinHelpers` [here](https://github.com/keep-network/tbtc.js/blob/8cedf3ee456fe5d1aa2cdbdb86d6532c8d6d6185/src/BitcoinHelpers.js#L19). We should also update the conditional above that ensures Bitcoin testnets are only used with web3 testnets.

liamzebedee: Yeah good point re: conditionals. 

A point for future - bcoin does export the same variety of networks here - https://sourcegraph.com/github.com/bcoin-org/bcoin@6beb54dddc1256cca379ad765b71fd5cdee427f3/-/blob/lib/protocol/networks.js. We could probably use those.

Shadowfiend: Nice find! I definitely want to reduce the number of hops between config and bcoin's networks. I am interested in maintaining independence from bcoin's structures though, as bcoin's networks aren't simple constants, they carry a lot of extra stuff. We might be able to build our constants directly off of their exposed networks by doing a `for (const [networkName, network] of Object.entries(networks)) ...` type thing. Though it also drives me bananas that they have `simnet`, `testnet`, and then `main` without a suffix. Would love to at least alias that or something.

Shadowfiend: Do we still need the `bcoin.set` here? All of the helpers in `BitcoinHelpers` explicitly pass the network info along from the config, I think.

PR #38, opened 2020-05-29
Deposit Submitter: Add tool for opening and funding deposits

Shadowfiend:
Meant for testing.

### DIFF
--- a/bin/tbtc-submitter.js
+++ b/bin/tbtc-submitter.js
@@ -1,0 +1,237 @@
+#!/usr/bin/env node --experimental-modules
+// @ts-check
+const description = `
+tbtc-submitter.js <ethereum-key-file> <bitcoin-wallet-file>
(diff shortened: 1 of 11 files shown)


bin/tbtc-submitter.js
```diff
@@ -0,0 +1,237 @@
...
+tbtc-submitter.js <ethereum-key-file> <bitcoin-wallet-file>
+    In default mode, creates a single tBTC deposit with the given Ethereum
+    key file and funds it using the Bitcoin wallet defined by the given
+    file.
+
+    -c,--continuous
+        This flag repeatedly creates tBTC deposits and funds them. Rather
+        than waiting for each deposit's funding transaction to fully
+        confirm, a new deposit opened once the previous deposit's signing
+        group becomes available.
```
Shadowfiend, line 13: Worth flagging, current code doesn't pay attention to a `-c` flag, and though it reads the two file parameters, it ignores them.

PR #43, opened 2020-06-02
Add tbtc.Constants

liamzebedee:
Constants class exposes the values of TBTCConstants contract from JS. 

Constants are retrieved using a BatchRequest, which improves performance and should result in fewer reqs for hosted providers like Infura.

### DIFF
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -1,0 +1,105 @@
+import TBTCConstantsJSON from "@keep-network/tbtc/artifacts/TBTCConstants.json"
+import EthereumHelpers from "./EthereumHelpers.js"
+import web3Utils from "web3-utils"
+const { toBN } = web3Utils
+
+/** @typedef { import("web3").default.Web3.eth.Contract } Contract */
+
+class Constants {
+  /**
+   * @param {TBTCConfig} config The config to use for this constants instance.
+   * @return {Constants} The TBTC constants.
+   */
+  static async withConfig(config) {
+    const { web3 } = config
+    const networkId = await web3.eth.net.getId()
+    const tbtcConstantsContract = EthereumHelpers.getDeployedContract(
+      TBTCConstantsJSON,
+      web3,
+      networkId
+    )
+
+    // BatchRequest makes a batch of Web3 RPC requests as one network payload.
+    // However, the API in web3@1 does not return a Promise with all results,
+    // as you might expect. So, we have to do this wizardry.
+    // See https://github.com/ethereum/web3.js/issues/1446.
+    const batch = new web3.BatchRequest()
+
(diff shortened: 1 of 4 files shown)


src/Constants.js
```diff
@@ -0,0 +1,105 @@
...
+        batch.add(request)
+      })
+    })
+
+    batch.execute()
+    const results = await Promise.all(calls)
+
+    const constants = {}
+    results.map(constantEntry => {
+      Object.assign(constants, constantEntry)
```
Shadowfiend, line 64: This will need to be `return` or just moved to one line without the curlies, thanks to JavaScript's magical “curlies remove the implied return” rule.

src/EthereumHelpers.js
```diff
@@ -151,10 +151,36 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
...
+    }
+    return deploymentInfo.address
+  }
+
+  const contract = new web3.eth.Contract(artifact.abi)
+  contract.options.address = lookupAddress(artifact)
+  contract.options.from = web3.eth.defaultAccount
+
+  return contract
+}
```
Shadowfiend, line 177: 👏👏👏👏

src/EthereumHelpers.js
```diff
@@ -151,10 +151,36 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
   }
 }
 
+/**
+ * Gets the Web3 Contract for a Truffle artifact and Web3 instance.
+ * @param {JSON} artifact
+ * @param {*} web3
+ * @param {*} networkId
```
Shadowfiend, line 158: Isn't this a `string`?

src/Constants.js
ironng: Should this be a `forEach` instead of `map`? Looks like the intent here is to update the `constants` object rather than doing anything with the new array `map` would generate. Or maybe we could use array `reduce`?

Shadowfiend: Derp, I didn't even look at the internal function. `forEach` or even just a standard `for` loop seems like the right move, good catch.

src/EthereumHelpers.js
```diff
@@ -155,7 +155,7 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
  * Gets the Web3 Contract for a Truffle artifact and Web3 instance.
  * @param {JSON} artifact
  * @param {*} web3
- * @param {*} networkId
+ * @param {string} networkId
```
Shadowfiend, line 158: … Actually I think this might be a `number`. We can always fix it when we run into it.

ironng: 🙈 😆

PR #44, opened 2020-06-03
Introduce examples/

liamzebedee:
`examples/` contains two snippets for opening a deposit and redemption.


examples/redeem.js
```diff
@@ -0,0 +1,65 @@
...
+  const engine = new ProviderEngine({ pollingInterval: 1000 })
+  engine.addProvider(
+    // For address 0x420ae5d973e58bc39822d9457bf8a02f127ed473.
+    new Subproviders.PrivateKeyWalletSubprovider(
+      "b6252e08d7a11ab15a4181774fdd58689b9892fe9fb07ab4f026df9791966990"
+    )
+  )
+  engine.addProvider(
+    new Subproviders.RPCSubprovider(
+      "https://:e18ef5ef295944928dd87411bc678f19@ropsten.infura.io/v3/59fb36a36fa4474b890c13dd30038be5"
```
Shadowfiend, line 17: Unfortunately I don't think this actually works, for here or perhaps for the deposit flow. We need a WebSocket subprovider, which requires some more fiddling around. I have some local code from launch time where I had to figure this out; will try and integrate it here soon.

PR #47, opened 2020-06-24
Expose redemption autosubmit state

ironng:
Redemption autoSubmit now returns an object with promises to various stages of the auto-submit process instead of a single promise. This now mirrors the format of the Deposit autoSubmit method.

### DIFF
--- a/src/Redemption.js
+++ b/src/Redemption.js
@@ -184,40 +184,38 @@ export default class Redemption {
             signedTransaction
           )
         }
-
-        return transaction
+        return transaction.transactionID
       }
     )
 
-    state.confirmations = state.broadcastTransactionID.then(async transaction => {
-      const requiredConfirmations = parseInt(
-        await this.deposit.factory.constantsContract.methods
-          .getTxProofDifficultyFactor()
-          .call()
-      )
+    state.confirmations = state.broadcastTransactionID.then(
+      async transactionID => {
+        const requiredConfirmations = parseInt(
+          await this.deposit.factory.constantsContract.methods
+            .getTxProofDifficultyFactor()
+            .call()
+        )
 
-      console.debug(
-        `Waiting for ${requiredConfirmations} confirmations for ` +
-          `Bitcoin transaction ${transaction.transactionID}...`
-      )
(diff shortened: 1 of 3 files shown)


src/Redemption.js
```diff
@@ -133,11 +133,31 @@ export default class Redemption {
...
-    if (this.autoSubmitPromise) {
-      return this.autoSubmitPromise
+    if (this.autoSubmittingState) {
+      return this.autoSubmittingState
     }
-    this.autoSubmitPromise = this.signedTransaction.then(
+
+    const state = (this.autoSubmittingState = {})
+
+    state.signedTransaction = this.signedTransaction.then(
```
Shadowfiend, line 160: I don't think this is quite right -> this is the broadcast transaction, not the signed transaction, and all it includes is the `transactionID` (if we get it from `findScript` it's actually a `FoundTransaction`, but if it comes from `broadcast` it isn't, so `transactionID` is all we can rely on). I think we can just rename to `broadcastTransaction` and be fine, or we can call it `broadcastTransactionID` and pull the ID out of the transaction when we `return` below.

src/Redemption.js
```diff
@@ -165,21 +185,31 @@ export default class Redemption {
...
+          .getTxProofDifficultyFactor()
+          .call()
+      )
 
+      console.debug(
+        `Waiting for ${requiredConfirmations} confirmations for ` +
+          `Bitcoin transaction ${transaction.transactionID}...`
+      )
+      await BitcoinHelpers.Transaction.waitForConfirmations(
+        transaction,
```
Shadowfiend, line 204: Heh. Really this should also take a `transactionID` rather than a `transaction`. Might be too deep a refactor right now though.

src/Redemption.js
ironng: You're right; we already got the signed transaction. Updated!

src/Redemption.js
ironng: Went for it. Updated!

Shadowfiend: :sunglasses:

PR #53, opened 2020-07-16
Adjust to rename in Deposit.signerFee

Shadowfiend:
The new name is signerFeeTbtc, which aligns with the tbtc.js method name
as well.

In draft until keep-network/tbtc#704, which implements this change, lands.


Shadowfiend:
The upstream PR has landed so this is ready for merge!

PR #55, opened 2020-07-24
Get network id from artifact

r-czajkowski:
Ref: https://github.com/keep-network/tbtc-dapp/pull/205

Added a function that returns the network id from tbtc artifact. Artifacts from `@keep-network/tbtc` for a given build only support a single network id.


r-czajkowski:
I think we can handle this in a similar way see https://github.com/keep-network/keep-core/pull/1629. We have 2 different workflows in our circle CI(keep_dev, keep_test, probably will be the 3rd workflow for mainnet) so we can pass some build args to docker image or something like that.

PR #58, opened 2020-07-28
Add additional methods to BitcoinHelper

lukasz-zimnoch:
Refs: https://github.com/keep-network/tbtc/issues/703

Implemented some additional methods exposed by the Electrum Client which will be needed by the end-to-end test script for testnet: https://github.com/keep-network/local-setup/pull/20

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -523,6 +523,55 @@ const BitcoinHelpers = {
 
       return transaction.toRaw().toString("hex")
     },
+    /**
+     * Finds all transactions containing unspent outputs received
+     * by the `bitcoinAddress`.
+     *
+     * @param {string} bitcoinAddress Bitcoin address to check.
+     *
+     * @return {Promise<Array>} A promise to the found array of transactions.
+     */
+    findAllUnspent: async function(bitcoinAddress) {
+      return await BitcoinHelpers.withElectrumClient(async electrumClient => {
+        const script = BitcoinHelpers.Address.toScript(bitcoinAddress)
+        return BitcoinHelpers.Transaction.findAllUnspentWithClient(electrumClient, script)
+      })
+    },
+    /**
+     * Gets the confirmed balance of the `bitcoinAddress`.
+     *
+     * @param {string} bitcoinAddress Bitcoin address to check.
+     *
+     * @return {Promise<Number>} A promise to the confirmed balance as satoshis.
+     */
+    getBalance: async function(bitcoinAddress) {
+      return await BitcoinHelpers.withElectrumClient(async electrumClient => {
+        const script = BitcoinHelpers.Address.toScript(bitcoinAddress)
+        return (await electrumClient.getBalanceOfScript(script)).confirmed
+      })
+    },
+    /**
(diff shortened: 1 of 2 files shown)


src/BitcoinHelpers.js
```diff
@@ -523,6 +523,55 @@ const BitcoinHelpers = {
...
+        const script = BitcoinHelpers.Address.toScript(bitcoinAddress)
+        return BitcoinHelpers.Transaction.findAllUnspentWithClient(electrumClient, script)
+      })
+    },
+    /**
+     * Gets the confirmed balance of the `bitcoinAddress`.
+     *
+     * @param {string} bitcoinAddress Bitcoin address to check.
+     *
+     * @return {Promise<Number>} A promise to the confirmed balance as satoshis.
```
Shadowfiend, line 545: in satoshis*

src/BitcoinHelpers.js
```diff
@@ -523,6 +523,55 @@ const BitcoinHelpers = {
...
+     * @param {number} [includeWithinBlocks=1] - The estimate is based on having
+     *        the transaction included within this number of blocks. Defaults to
+     *        trying to include the transaction within 1 block.
+     *
+     * @return {Promise<number>} The estimated fee to execute the provided
+     *         transaction. Fails if the fee cannot be estimated.
+     */
+    estimateFeePerKb: async function(includeWithinBlocks) {
+      return BitcoinHelpers.withElectrumClient(async electrumClient => {
+        const feePerKb = await electrumClient.getFeeEstimate(includeWithinBlocks || 1)
```
Shadowfiend, line 566: Worth noting, this was always returning 0 in my deposit submitter experiments, which is why I ended up always using the minimum relay fee in [the other PR's implementation](https://github.com/keep-network/tbtc.js/pull/38/files#diff-d8016928a1a8bbd777a2a7d57dc8cbbeR181).

lukasz-zimnoch:
Thanks for a turbo review @Shadowfiend!. Yep, it's still a draft because I'm still testing everything and there is still a possibility I will change something. Once I'm done with testing I'll make it ready for review and ping you.

src/BitcoinHelpers.js
lukasz-zimnoch: Yep, I have the same. I thought there will be a difference if I use it against BTC testnet but nope. I will remove this method and use the default ones or the minimum relay fee in my test script.

lukasz-zimnoch:
@Shadowfiend I'm ready here.

src/BitcoinHelpers.js
```diff
@@ -542,36 +542,14 @@ const BitcoinHelpers = {
...
-      return BitcoinHelpers.withElectrumClient(async electrumClient => {
-        const feePerKb = await electrumClient.getFeeEstimate(includeWithinBlocks || 1)
-
-        if (feePerKb < 0) {
-          throw new Error(`Fee cannot be estimated; Electrum returned ${feePerKb}.`)
-        }
-
-        return feePerKb
-      })
-    },
```
Shadowfiend: I'm definitely interested in making this work in the future, btw. Need to see what about our node is making the server side of this fall over.

PR #59, opened 2020-07-28
Notification Station: Pull in deposit API changes, plus fix two outstanding issues

### LINKED ISSUE #20: Use `handleRevert` for transaction calls
`handleRevert` alleviates some of the work of getting any error messages from failing transactions. 

https://web3js.readthedocs.io/en/v1.2.5/web3-eth-contract.html#handlerevert

### LINKED ISSUE #56: Adapt interface to Deposit function name updates
keep-network/tbtc#695 made a variety of tweaks to the public API the Deposit contract exposes. tbtc.js needs to be aligned to those to remain compatible.

### LINKED ISSUE #57: toNumber usage in errors with large BNs can lead to exceptions
The below code:

https://github.com/keep-network/tbtc.js/blob/8e3f8c796b17e50b68d569e918f75746bb9896eb/src/Deposit.js#L668-L670

Uses `toNumber` to interpolate a `BN` into a string. For numbers that are in a `BN` that are too large to represent as standard JS numbers, this will cause an exception constructing the error.

These instances should use `toString` instead.

Originally manifested as keep-network/tbtc-dapp#245 .

Shadowfiend:
Each commit handles a different issue. It turns out the current tbtc.js doesn't
have a lot of exposure to the changed APIs from keep-network/tbtc#695, so
that fix was small enough to roll in a few miscellaneous ones as well.

Of note, #25 likely will need to be revised more for the aforementioned API
changes; that can be done in-PR, however.

Closes #20, closes #56, closes #57.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -392,7 +392,7 @@ export default class Deposit {
    * @return {DepositStates} The current state of the deposit.
    */
   async getCurrentState() {
-    return parseInt(await this.contract.methods.getCurrentState().call())
+    return parseInt(await this.contract.methods.currentState().call())
   }
 
   async getTDT() /* : Promise<TBTCDepositToken>*/ {
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -287,11 +287,13 @@ export default class Deposit {
     const web3 = factory.config.web3
     const contract = new web3.eth.Contract(DepositJSON.abi, depositAddress)
     contract.options.from = web3.eth.defaultAccount
+    contract.options.handleRevert = true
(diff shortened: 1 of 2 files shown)


src/Deposit.js
```diff
@@ -287,11 +287,13 @@ export default class Deposit {
     const web3 = factory.config.web3
     const contract = new web3.eth.Contract(DepositJSON.abi, depositAddress)
     contract.options.from = web3.eth.defaultAccount
+    contract.options.handleRevert = true
```
ironng, line 290: 😍

PR #63, opened 2020-08-10
Replace obsolete getSatoshiLotSize usages

lukasz-zimnoch:
`getSatoshiLotSize` method was renamed to `getLotSizeSatoshis`. However, there was two places which still used the old name and caused errors. Here we fix this problem.


pdyraga:
Reference PR: https://github.com/keep-network/tbtc.js/pull/25

PR #66, opened 2020-08-13
Expose confirmations

### LINKED ISSUE #30: Expose confirmation events and current confirmation count during funding/redemption
Right now, though we have promises to the states before and after the right number of confirmations are received, there is no direct access to the current number of seen confirmations or to update that number as new confirmations come in.

ironng:
Expose `onReceivedConfirmation` events for both deposit and redemption flows. They take a handler that provides an object with the `transactionID` and current number of `confirmations` as its parameter. 

Closes #30

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -355,9 +355,9 @@ const BitcoinHelpers = {
             transactionID
           )
 
-          typeof onReceivedConfirmation === "function" &&
-            confirmations &&
+          if (typeof onReceivedConfirmation === "function" && confirmations) {
             onReceivedConfirmation({ transactionID, confirmations })
+          }
 
           if (confirmations >= requiredConfirmations) {
             return confirmations
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -338,7 +338,7 @@ const BitcoinHelpers = {
      * @param {string} transactionID A hex Bitcoin transaction id hash.
      * @param {number} requiredConfirmations The number of required
      *        confirmations to wait before returning.
-     * @param {function} onReceivedConfirmation A callback that fires when a
+     * @param {function} [onReceivedConfirmation] A callback that fires when a
      *        confirmation is seen
      *
      * @return {Promise<number>} A promise to the final number of confirmations
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -338,18 +338,30 @@ const BitcoinHelpers = {
      * @param {string} transactionID A hex Bitcoin transaction id hash.
      * @param {number} requiredConfirmations The number of required
      *        confirmations to wait before returning.
+     * @param {function} onReceivedConfirmation A callback that fires when a
+     *        confirmation is seen
      *
      * @return {Promise<number>} A promise to the final number of confirmations
      *         observed that was at least equal to the required confirmations.
      */
-    waitForConfirmations: async function(transactionID, requiredConfirmations) {
+    waitForConfirmations: async function(
+      transactionID,
+      requiredConfirmations,
+      onReceivedConfirmation
+    ) {
       return BitcoinHelpers.withElectrumClient(async electrumClient => {
         const checkConfirmations = async function() {
-          return await BitcoinHelpers.Transaction.checkForConfirmationsWithClient(
-            electrumClient,
-            transactionID,
-            requiredConfirmations
+          const { confirmations } = await electrumClient.getTransaction(
+            transactionID
           )
+
+          typeof onReceivedConfirmation === "function" &&
+            confirmations &&
+            onReceivedConfirmation({ transactionID, confirmations })
+
+          if (confirmations >= requiredConfirmations) {
+            return confirmations
+          }
         }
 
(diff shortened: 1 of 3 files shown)


src/BitcoinHelpers.js
```diff
@@ -338,18 +338,30 @@ const BitcoinHelpers = {
      * @param {string} transactionID A hex Bitcoin transaction id hash.
      * @param {number} requiredConfirmations The number of required
      *        confirmations to wait before returning.
+     * @param {function} onReceivedConfirmation A callback that fires when a
```
Shadowfiend, line 341: Let's mark this optional (by putting the name in square braces).

src/BitcoinHelpers.js
```diff
@@ -338,18 +338,30 @@ const BitcoinHelpers = {
...
-            electrumClient,
-            transactionID,
-            requiredConfirmations
+          const { confirmations } = await electrumClient.getTransaction(
+            transactionID
           )
+
+          typeof onReceivedConfirmation === "function" &&
+            confirmations &&
+            onReceivedConfirmation({ transactionID, confirmations })
```
Shadowfiend, line 360: My heart wants this, but my maintainer suggests we upgrade this to a proper if statement 😁

src/BitcoinHelpers.js
```diff
@@ -338,18 +338,30 @@ const BitcoinHelpers = {
...
      */
-    waitForConfirmations: async function(transactionID, requiredConfirmations) {
+    waitForConfirmations: async function(
+      transactionID,
+      requiredConfirmations,
+      onReceivedConfirmation
+    ) {
       return BitcoinHelpers.withElectrumClient(async electrumClient => {
         const checkConfirmations = async function() {
-          return await BitcoinHelpers.Transaction.checkForConfirmationsWithClient(
```
Shadowfiend, line 348: Are we using `checkForConfirmationsWithClient` elsewhere in the code? If not, we may want to just drop it from this file---it seems a bit  vestigial.

src/Deposit.js
```diff
@@ -523,6 +533,21 @@ export default class Deposit {
     // Bitcoin txHash; no verification initially.
   }
 
+  /**
+   * Registers a handler for notification when the Bitcoin funding transaction
+   * has received a confirmation
+   *
+   * @param {OnReceivedConfirmationHandler} onReceivedConfirmationHandler
+   *        A handler that passes an object with the transactionID and
+   *        confirmations as its parameter
```
Shadowfiend, line 542: Periods on both of these sentences please :)

I would also say the handler receives rather than passes those parameters, from the caller's perspective.

src/Deposit.js
```diff
@@ -523,6 +533,21 @@ export default class Deposit {
...
+   */
+  onReceivedConfirmation(onReceivedConfirmationHandler) {
```
Shadowfiend, line 544: Thoughts on clarifying the terminology in this class as "funding transaction confirmation" for the docs and function and variable names? It's long winded, but since the `Deposit` class also touches redemption, being clear will be useful I think.

src/BitcoinHelpers.js
ironng: 😆 I feel you. Updated!

src/BitcoinHelpers.js
ironng: Good call! It was still in use in `checkForConfirmations`. Refactored to just use that instead.

src/Deposit.js
ironng: Addressed!

src/Deposit.js
ironng: I was debating that myself but went for the shorter name. I figured we could talk about it in review :D

Renamed to `onReceivedFundingConfirmation` since we a getter called `fundingConfirmation` and also it is a bit shorter than `onReceivedFundingTransactionConfirmation`. I guess we could go with `onFundingTransactionConfirmation` or `onFundingConfirmation` maybe too? Let me know your preference. I think it is also little tricky to name because it's when the transaction receives a single confirmation vs all required confirmations.

src/BitcoinHelpers.js
```diff
@@ -306,29 +306,38 @@ const BitcoinHelpers = {
...
       return BitcoinHelpers.withElectrumClient(async electrumClient => {
-        return await BitcoinHelpers.Transaction.checkForConfirmationsWithClient(
-          electrumClient,
-          id,
-          requiredConfirmations
+        const { confirmations } = await electrumClient.getTransaction(
+          transactionID
         )
+
+        if (typeof onReceivedConfirmation === "function" && confirmations) {
```
Shadowfiend, line 334: Oof, I think the issue here is an `onReceivedConfirmation` parameter doesn't really make sense for a function that's just a one-shot true/false like `checkForConfirmations`. I think it probably makes more sense to drop the `onReceivedConfirmation` bit from here and just duplicate the first 3 lines of the callback in `waitForConfirmations`. What do you think?

ironng: Ah right. I think actually that's why I didn't do this before lol. Will switch it!

ironng: Done!

PR #67, opened 2020-08-14
Retry in case of retrieveSignerPubkey error

lukasz-zimnoch:
The `retrieveSignerPubkey` transaction can fail if they occur immediately after receiving the `PublicKeyPublished` event and a load-balanced Ethereum client is used. This is because the contract state is propagated with a delay across all Ethereum nodes. There can be a situation when the `PublicKeyPublished` is received from one node but the `retrieveSignerPubkey` is executed against another one which isn't aware of the public key yet and will revert the transaction. To prevent breaking the whole deposit flow, we introduce a retry mechanism.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -979,8 +979,11 @@ export default class Deposit {
       `Waiting for deposit ${this.address} to retrieve public key...`
     )
     // Ask the deposit to fetch and store the signer pubkey.
-    const pubkeyTransaction = await EthereumHelpers.sendSafely(
-      this.contract.methods.retrieveSignerPubkey()
+    const pubkeyTransaction = await EthereumHelpers.sendSafelyRetryable(
+      this.contract.methods.retrieveSignerPubkey(),
+      {},
+      false,
+      5
     )
 
     console.debug(`Found public key for deposit ${this.address}...`)
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -978,26 +978,10 @@ export default class Deposit {
     console.debug(
       `Waiting for deposit ${this.address} to retrieve public key...`
     )
-
-    let pubkeyTransaction
-
-    for (;;) {
-      try {
-        // Ask the deposit to fetch and store the signer pubkey.
-        console.debug(`Starting public key retrieval attempt...`)
-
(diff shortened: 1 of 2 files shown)


src/Deposit.js
```diff
@@ -978,10 +978,26 @@ export default class Deposit {
...
       `Waiting for deposit ${this.address} to retrieve public key...`
     )
-    // Ask the deposit to fetch and store the signer pubkey.
-    const pubkeyTransaction = await EthereumHelpers.sendSafely(
-      this.contract.methods.retrieveSignerPubkey()
-    )
+
+    let pubkeyTransaction
+
+    for (;;) {
```
Shadowfiend, line 984: I don't know about retrying infinitely here… It seems like a bit much, no? This could be a legitimate failure, after all.

I'm wondering if it might make sense to build the retry logic into `sendSafely`, with an optional parameter to specify the number of retries and an exponential backoff on the retry time. Thoughts?

lukasz-zimnoch: > I don't know about retrying infinitely here… It seems like a bit much, no? This could be a legitimate failure, after all.

Yeah, you're right.

> I'm wondering if it might make sense to build the retry logic into sendSafely, with an optional parameter to specify the number of retries and an exponential backoff on the retry time. Thoughts?

Yeah sure, we can make it generic.

lukasz-zimnoch:
Ready for the next round @Shadowfiend

src/EthereumHelpers.js
```diff
@@ -151,6 +151,54 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
...
+
+      console.debug(
+        `attempt ${attempt} failed: ${exception}; ` +
+          `retrying after ${waitMillis} milliseconds`
+      )
+
+      await new Promise(resolve => setTimeout(resolve, waitMillis))
+    }
+  }
+}
```
Shadowfiend, line 200: Honestly I would say that having `sendSafely` retry 2 or 3 times by default, rather than adding another method for it, doesn't seem crazy… But there's some more massaging I need to do around the whole `sendSafely` situation anyway, so let's stick to this for now and then we can reconcile later.

Issue #65, opened 2020-08-12
Complete typing and enable ts-check across the repo


Shadowfiend:
It's time! We've had partial typing in JSDoc across the repository, but haven't finished cleaning up type errors. The goal here is to get to a point where the whole project can be successfully type-checked with TypeScript (with a little help from Web3's latest typing updates).

Issue #32, opened 2020-05-16
Web3 must set a defaultAccount


dmihal:
When trying to run through a deposit, I ran into an issue where transactions can't be sent because `web3.eth.defaultAccount` wasn't set.

I don't think its a common practice for defs to set this value, at least I've never used it.

Options to address this would be:

* Update the documentation, making it clear that this value should be set
* Query web3.eth.getAccounts() and use account[0]

liamzebedee:
@Shadowfiend what do you think? I'm comfortable with option 2.

Shadowfiend:
I think the move here is to provide the ability to specify an account in config, fall back on `defaultAccount`, and throw if neither is set rather than implicitly guessing---particularly when there's already a mechanism to spare us the guess in `defaultAccount`.

PR #70, opened 2020-08-17
Bit Dusty Corners: Improve typing in Bitcoin and Ethereum helpers

Shadowfiend:
This is a fairly comprehensive rework for correctness of typing across all
Bitcoin and Ethereum helper files. It doesn't touch any of the direct tBTC
files, which will get their own PR. That second PR will also land the jsconfig
update that flags all of these errors, and may include a pre-commit hook
for TypeScript type checking.

See #65 .

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -23,9 +23,20 @@ const BitcoinNetwork = {
 }
 
 /**
- * Found transaction details.
- * @typedef FoundTransaction
- * @type {Object}
+ * Partial details about a transaction; same fields as `Transaction`, but only
+ * the transaction ID is guaranteed to be present in a function that returns
+ * this type.
+ *
+ * @typedef {object} PartialTransactionInBlock
+ * @property {string} transactionID Transaction ID.
+ * @property {number} [outputPosition] Position of output in the transaction.
+ * @property {number} [value] Value of the output (satoshis).
+ */
+
+/**
+ * Complete details about a transaction in a block.
+ *
+ * @typedef {object} TransactionInBlock
  * @property {string} transactionID Transaction ID.
  * @property {number} outputPosition Position of output in the transaction.
  * @property {number} value Value of the output (satoshis).
@@ -243,7 +254,7 @@ const BitcoinHelpers = {
      * @param {number} expectedValue The expected value of the transaction
      *        to fetch.
      *
-     * @return {Promise<FoundTransaction>} A promise to an object of
+     * @return {Promise<TransactionInBlock>} A promise to an object of
      *         transactionID, outputPosition, and value, that resolves with
      *         either null if such a transaction could not be found, or the
      *         information about the transaction that was found.
@@ -263,7 +274,7 @@ const BitcoinHelpers = {
      * @param {number} expectedValue The expected value of the transaction
      *        to fetch.
      *
-     * @return {Promise<FoundTransaction>} A promise to an object of
+     * @return {Promise<TransactionInBlock>} A promise to an object of
      *         transactionID, outputPosition, and value, that resolves with
      *         either null if such a transaction could not be found, or the
      *         information about the transaction that was found.
@@ -284,7 +295,7 @@ const BitcoinHelpers = {
      * @param {string} bitcoinAddress Bitcoin address to watch.
      * @param {number} expectedValue The expected value to watch for.
      *
-     * @return {Promise<FoundTransaction>} A promise to the found
+     * @return {Promise<TransactionInBlock>} A promise to the found
      *         transaction once it is seen on the chain.
      */
     findOrWaitFor: async function(bitcoinAddress, expectedValue) {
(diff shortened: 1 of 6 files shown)


src/BitcoinHelpers.js
```diff
@@ -45,12 +58,25 @@ const BitcoinNetwork = {
...
- * @typedef {Object} SPVProof
- * @extends {Proof}
- * @property {ParsedTransaction} parsedTransaction Parsed transaction with
+ * @typedef {object} SPVProofInfo
+ * @property {HexTransactionFields} parsedTransaction Parsed transaction with
  *           additional data useful in submitting SPV proofs, stored as buffers.
  */
 
+/**
+ * @typedef {Proof & SPVProofInfo} SPVProof
```
Shadowfiend, line 67: The previous way didn't correctly articulate that `SPVProof` should have everything in `Proof` + the `parsedTransaction` field.

src/BitcoinHelpers.js
```diff
@@ -476,7 +506,10 @@ const BitcoinHelpers = {
       // Combine witness
       let signedTransaction
       try {
-        signedTransaction = bcoin.TX.fromRaw(unsignedTransaction, "hex").clone()
+        signedTransaction = bcoin.TX.fromRaw(
+          Buffer.from(unsignedTransaction, "hex"),
```
Shadowfiend, line 510: We got lucky here that bcoin could take a hex string in `fromRaw` despite declaring that it always takes a `Buffer`. This does the buffer conversion on our end to conform with the declared API.

src/BitcoinHelpers.js
```diff
@@ -45,12 +58,25 @@ const BitcoinNetwork = {
  */
 
 /**
- * @typedef {Object} SPVProof
- * @extends {Proof}
- * @property {ParsedTransaction} parsedTransaction Parsed transaction with
+ * @typedef {object} SPVProofInfo
+ * @property {HexTransactionFields} parsedTransaction Parsed transaction with
```
ironng, line 62: Is it worth renaming the property to `hexTransactionFields` since the `ParsedTransaction` type no longer exists? Also fine keeping as is.

src/BitcoinHelpers.js
```diff
@@ -45,12 +58,25 @@ const BitcoinNetwork = {
...
+/**
+ * @typedef {Proof & SPVProofInfo} SPVProof
+ */
+
+/**
+ * A handler that is notified whenever a new confirmation arrives for a given
+ * transaction id.
+ *
+ * @callback OnReceivedConfirmationHandler
+ * @param {{ transactionID: string, confirmations: number }} confirmationInfo
```
ironng, line 75: ✨

src/EthereumHelpers.js
```diff
@@ -129,35 +158,21 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
...
-
-      if (forceSend) {
-        return boundContractMethod.send(sendParams)
-      } else {
-        // If we weren't able to get a better error from `call`, throw the
-        // original exception.
-        throw exception
-      }
+      return boundContractMethod.send({
+        from: "", // FIXME Need systemic handling of default from address.
```
ironng, line 173: Should we open an issue for the `FIXME`s?

src/EthereumHelpers.js
```diff
@@ -129,35 +158,21 @@ async function sendSafely(boundContractMethod, sendParams, forceSend) {
...
-    // If we're not forcibly sending, try to resolve the true error by using
-    // `call`.
+    // For an always failing transaction, if forceSend is set, send it anyway.
     if (
       exception.message &&
-      exception.message.match(/always failing transaction/)
+      exception.message.match(/always failing transaction/) &&
+      forceSend
     ) {
-      let callResult
```
ironng, line 142: Just curious about why we're dropping this logic? Is it because it's not actually doing anything (i.e. "FIXME Something more is needed here to properly resolve this error...)?

src/lib/ElectrumClient.js
```diff
@@ -292,38 +292,26 @@ export default class Client {
...
+   * Get proof of transaction inclusion in the block.
+   *
    * @param {string} txHash Hash of a transaction.
-   * @param {number} blockHeight Height of the block where transaction was confirmed.
-   * @return {string} Transaction inclusion proof in hexadecimal form.
+   * @param {number} blockHeight Height of the block where transaction was
+   *        confirmed.
+   * @return {Promise<any>} Transaction inclusion proof in hexadecimal form.
    */
-  // TODO: We should move this function to BitcoinSPV.js file as it's strictly
```
ironng, line 302: 👏 Yay for sniping this

src/BitcoinHelpers.js
Shadowfiend: I wanted to avoid logic changes where possible, but I think we'll be able to refactor this confidently with VS Code tooling after all of this stuff lands.

src/EthereumHelpers.js
Shadowfiend: We have one! It's #32 .

src/EthereumHelpers.js
Shadowfiend: Ah yes, I forgot to flag this at the PR level! Short version here is---it wasn't doing anything. `handleRevert` bubbles the error up without this extra dance, and before `handleRevert` was set, this didn't actually do what it was supposed to---though I'm not sure why :/

ironng: Aha. Thanks for clarifying!

PR #71, opened 2020-08-17
Type-sy: Just a few types to drink

Shadowfiend:
This PR cleans up and refines types across the public tbtc.js API
surface, and adds a tsc-based linting step that checks types before
commits go through. Right now only the src/ directory is linted, but
the examples and bin directories will be close behind.

Additionally, the new `jsconfig.json` should be enough to have VS
Code and other JS language server users apply proper type-checking
throughout the project.

Draft until #70 lands.

See #65 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,10 @@
     "test": "mocha --timeout 10000",
     "lint": "npm run lint:js",
     "lint:fix": "npm run lint:fix:js",
-    "lint:js": "eslint .",
-    "lint:fix:js": "eslint --fix ."
+    "lint:js": "npm run lint:js:eslint && npm run lint:js:types",
+    "lint:fix:js": "eslint --fix .",
+    "lint:js:eslint": "eslint .",
+    "lint:js:types": "!(npx tsc --allowJs --noEmit $(cat jsconfig.json | jq -r '.compilerOptions | to_entries | map([\"--\\(.key)\",.value]) | flatten | join(\" \")') src/**.js | grep \"^\\(src\\|bin\\|test\\|examples\\)/\") # comment any other passed arguments"
   },
   "author": "Antonio Salazar Cardozo <antonio@thesis.co>",
   "license": "MIT",
@@ -47,6 +49,7 @@
     "eslint-config-keep": "git+https://github.com/keep-network/eslint-config-keep.git#0.3.0",
     "prettier": "^1.19.1",
     "fs": "0.0.1-security",
-    "mocha": "^6.2.0"
+    "mocha": "^6.2.0",
+    "typescript": "3.4.3"
   }
 }
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -168,7 +168,7 @@ export class DepositFactory {
       const contract = EthereumHelpers.getDeployedContract(
         artifact,
         web3,
-        networkId
+        networkId.toString()
       )
       this[propertyName] = contract
     })
@@ -240,8 +240,8 @@ export class DepositFactory {
       await this.systemContract.methods.getNewDepositFeeEstimate().call()
     )
 
-    const accountBalance = await this.config.web3.eth.getBalance(
-      this.config.web3.eth.defaultAccount
+    const accountBalance = toBN(
+      await this.config.web3.eth.getBalance(this.config.web3.eth.defaultAccount)
     )
 
     if (creationCost.lt(accountBalance)) {
@@ -410,7 +410,10 @@ export default class Deposit {
         console.debug(
           `Monitoring Bitcoin for transaction to address ${address}...`
         )
-        return BitcoinHelpers.Transaction.findOrWaitFor(address, expectedValue)
+        return BitcoinHelpers.Transaction.findOrWaitFor(
+          address,
+          expectedValue.toNumber()
+        )
       }))
   }
 
@@ -661,7 +664,7 @@ export default class Deposit {
     )
     const proofArgs = await this.constructFundingProof(
       tx,
-      parseInt(requiredConfirmations)
+      requiredConfirmations
     )
     proofArgs.unshift(this.address)
 
@@ -686,7 +689,7 @@ export default class Deposit {
       "Transfer"
     )
 
-    return toBN(transferEvent.value).div(toBN(10).pow(18))
+    return toBN(transferEvent.value).div(toBN(10).pow(toBN(18)))
   }
 
   /**
@@ -986,7 +989,7 @@ export default class Deposit {
         )
         console.debug(`Minted`, transferEvent.value, `TBTC.`)
 
-        return toBN(transferEvent.value).div(toBN(10).pow(18))
+        return toBN(transferEvent.value).div(toBN(10).pow(toBN(18)))
       }
     )
 
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -889,9 +889,10 @@ export default class Deposit {
    * @typedef {Object} AutoSubmitState
    * @prop {Promise<BitcoinTransaction>} fundingTransaction
    * @prop {Promise<{ transaction: BitcoinTransaction, requiredConfirmations: Number }>} fundingConfirmations
-   * @prop {Promise<TransactionReceipt>} proofTransaction
-   * @prop {Promise<number>} mintedTBTC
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -22,8 +22,10 @@
     "test": "mocha --timeout 10000",
     "lint": "npm run lint:js",
     "lint:fix": "npm run lint:fix:js",
-    "lint:js": "eslint .",
-    "lint:fix:js": "eslint --fix ."
+    "lint:js": "npm run lint:js:eslint && npm run lint:js:types",
+    "lint:fix:js": "eslint --fix .",
+    "lint:js:eslint": "eslint .",
+    "lint:js:types": "!(npx tsc --allowJs --noEmit $(cat jsconfig.json | jq -r '.compilerOptions | to_entries | map([\"--\\(.key)\",.value]) | flatten | join(\" \")') src/**.js | grep \"^\\(src\\|bin\\|test\\|examples\\)/\")"
```
Shadowfiend, line 28: This nasty dance lets us do two things:
- Pull most of the compiler settings from `jsconfig.json`.
- Only look at errors from our files (there's no way to get the type checking we want and have TypeScript only report errors in our files---`node_modules` either reports errors or the type information from modules isn't used for our files).
- The `!` up front inverts the exit code from `grep`. `grep` returns `0` when it finds a match and `1` when it finds no match, but we want the command to *fail* when `grep` finds a match and *pass* when it finds no match, so we invert the exit code.

src/Deposit.js
```diff
@@ -1,13 +1,16 @@
-import EventEmitter from "events"
+import { EventEmitter } from "events"
```
Shadowfiend, line 1: This is filed under “wait, did that even work?”

src/Deposit.js
```diff
@@ -224,8 +243,8 @@ export class DepositFactory {
       await this.systemContract.methods.getNewDepositFeeEstimate().call()
     )
 
-    const accountBalance = await this.config.web3.eth.getBalance(
-      this.config.web3.eth.defaultAccount
+    const accountBalance = toBN(
+      await this.config.web3.eth.getBalance(this.config.web3.eth.defaultAccount)
```
Shadowfiend, line 247: This is filed under “that didn't even work, but it failed in a way that allowed things to continue as expected”.

src/Deposit.js
```diff
@@ -392,7 +413,10 @@ export default class Deposit {
         console.debug(
           `Monitoring Bitcoin for transaction to address ${address}...`
         )
-        return BitcoinHelpers.Transaction.findOrWaitFor(address, expectedValue)
+        return BitcoinHelpers.Transaction.findOrWaitFor(
+          address,
+          expectedValue.toNumber()
```
Shadowfiend, line 418: This worked before because JavaScript will coerce a value into a number when needed; this lets us be more explicit, however.

src/Deposit.js
```diff
@@ -643,9 +667,8 @@ export default class Deposit {
     )
     const proofArgs = await this.constructFundingProof(
       tx,
-      parseInt(requiredConfirmations)
+      requiredConfirmations
```
Shadowfiend, line 670: This worked because `parseInt` doesn't do anything when passed a number, but it was redundant.

src/Deposit.js
```diff
@@ -643,9 +667,8 @@ export default class Deposit {
...
     )
-    proofArgs.unshift(this.address)
```
Shadowfiend, line 648: This is again filed under “wait, did this even work?”

src/Deposit.js
```diff
@@ -668,7 +691,7 @@ export default class Deposit {
       "Transfer"
     )
 
-    return toBN(transferEvent.value).div(toBN(10).pow(18))
+    return toBN(transferEvent.value).div(toBN(10).pow(toBN(18)))
```
Shadowfiend, line 694: Same as above, this should be throwing an exception.

src/Deposit.js
```diff
@@ -1231,9 +1259,10 @@ export default class Deposit {
    * Purchases the signer bonds and closes the liquidation auction.
    */
   async purchaseSignerBondsAtAuction() {
+    // FIXME Need systemic handling of default from address.
     const owner = this.factory.config.web3.eth.defaultAccount
     const allowance = await this.factory.tokenContract.methods
-      .allowance(owner, deposit.address)
+      .allowance(owner, this.address)
```
Shadowfiend, line 1265: This was broken, we just weren't invoking the containing function.

Shadowfiend:
This one needs a test drive against a local setup---a few places make actual logic changes that should be fixing bugs, but I haven't had a chance to confirm yet.

src/Deposit.js
ironng: 😆 It actually did!

src/Deposit.js
```diff
@@ -294,66 +313,68 @@ export default class Deposit {
...
+      /** @type {TruffleArtifact} */ (DepositJSON).abi,
+      depositAddress
+    )
+    const keepContract = EthereumHelpers.buildContract(
+      web3,
+      /** @type {TruffleArtifact} */ (BondedECDSAKeepJSON).abi,
       keepAddress
     )
-    keepContract.options.from = web3.eth.defaultAccount
-    keepContract.options.handleRevert = true
```
ironng, line 305: Curious why we're dropping setting these options for `contract` and `keepContract?

src/Deposit.js
```diff
@@ -898,30 +924,30 @@ export default class Deposit {
...
-      async ({ transaction, requiredConfirmations }) => {
-        console.debug(
-          `Submitting funding proof to deposit ${this.address} for ` +
-            `Bitcoin transaction ${transaction.transactionID}...`
-        )
-        const proofArgs = await this.constructFundingProof(
-          transaction,
-          requiredConfirmations
-        )
+    this.autoSubmittingState = {
```
ironng, line 927: 😍

src/Deposit.js
Shadowfiend: 😅 the follow-up question is immediately: will it still work? <_<

src/Deposit.js
Shadowfiend: We aren't! I introduced `EthereumHelpers.buildContract` in the previous PR, which replaces our manual `new web3.eth.Contract` invocations and takes care of setting this stuff in one place. This change moves `keepContract` to use that function, so we don't have to do that explicitly here :tada:

src/Deposit.js
ironng: Fair point! :)

src/Deposit.js
ironng: 😎 Very cool. 😅 I do remember seeing that new method in the last PR, but I missed it anyway lol 🙃

src/Deposit.js
ironng: Running the local test, this new change is causing an error for the next line:
```
{
    "type": "error",
    "timestamp": "Tue, 18 Aug 2020 22:35:16 GMT",
    "arguments": {
      "0": {
        "message": "Insufficient balance 1000000000000000000000000000000124273889123276000000000 to open deposit (required: 1164000000000000).",
        "originalStack": [
          "Error: Insufficient balance 1000000000000000000000000000000124273889123276000000000 to open deposit (required: 1164000000000000).",
          "at DepositFactory.createNewDepositContract (http://localhost:3000/static/js/main.chunk.js:1082:13)",
          "at async Function.forLotSize (http://localhost:3000/static/js/main.chunk.js:1124:9)"
        ],
        "error": {}
      }
    },
```

This happens when I select the lot size of 0.01 BTC

Shadowfiend: Oh goodness.

Yeah that check is inverted 😂/:sob: Will change it to `.gte`.

Shadowfiend:
Uh-oh, I did something wrong here. Hold please.

src/Deposit.js
ironng: I think maybe the check itself is okay - `creationCost.lt(accountBalance)` but I think the param for `.lt` needs to be a plain JS number instead of a BN?

Shadowfiend:
Ok, I think we're good here now.

Shadowfiend:
(I do need to rebase the following PRs before taking them out of draft though. That'll have to be tomorrow.)

PR #74, opened 2020-08-18
Explicitly strict: Upgrade types to support strict null checks and no implicit any

Shadowfiend:
TypeScript provides a couple of stricter options, strict null checks and no implicit
any. These are described in the [TypeScript compiler docs](https://www.typescriptlang.org/docs/handbook/compiler-options.html)), but brief descriptions
are below.

By default, TypeScript allows a variable of most types to be assigned `null`. Strict
null checks disables this, and requires that you explicitly flag when you expect that
a value might be null---and requires that you handle that possibility in code that
consumes that type. This required quite a few logic adjustments. In most cases, this
PR adds throws when we would only get a null in a misconfigured invocation. In other
cases, a default is introduced.

Turning on strict null checks also required adjusting how types are declared for things
that were being looked up asynchronously. The biggest change is in how the contract
objects that `Deposit` and `DepositFactory` track are referenced. These now have
wrapper getter functions that take care of going from `Contract | undefined` to a known
`Contract` type, throwing if that contract wasn't properly initialized. A similar shift needed
to happen in `Constants`, where values that are looked up at initialization needed to be
declared as available on the `Constants` type. In that case, an intermediary type is introduced
that handles declaring those properties.

By default, TypeScript allows any variable whose type can't be inferred to implicitly
have the `any` type, which is more or less not checked for any correctness. No implicit
any disables this, and requires that a variable's type either be able to be inferred, or
that an explicit type be declared for it. This sprouts a lot of additional type annotations
throughout the codebase, including `BitcoinTxParser`.

In draft until #71 lands.

See #65.


Shadowfiend:
Ohhhh Lordy. I need to redo this whole thing after my fun rebase adventure 😭

Shadowfiend:
And the follow up. Looollll I can git I swears.

PR #77, opened 2020-08-19
Explicitly Strict: Upgrade types to support strict null checks and no implicit any

Shadowfiend:
TypeScript provides a couple of stricter options, strict null checks and no implicit
any. These are described in the TypeScript compiler docs), but brief descriptions
are below.

By default, TypeScript allows a variable of most types to be assigned null. Strict
null checks disables this, and requires that you explicitly flag when you expect that
a value might be null---and requires that you handle that possibility in code that
consumes that type. This required quite a few logic adjustments. In most cases, this
PR adds throws when we would only get a null in a misconfigured invocation. In other
cases, a default is introduced.

Turning on strict null checks also required adjusting how types are declared for things
that were being looked up asynchronously. The biggest change is in how the contract
objects that Deposit and DepositFactory track are referenced. These now have
wrapper getter functions that take care of going from Contract | undefined to a known
Contract type, throwing if that contract wasn't properly initialized. A similar shift needed
to happen in Constants, where values that are looked up at initialization needed to be
declared as available on the Constants type. In that case, an intermediary type is introduced
that handles declaring those properties.

By default, TypeScript allows any variable whose type can't be inferred to implicitly
have the any type, which is more or less not checked for any correctness. No implicit
any disables this, and requires that a variable's type either be able to be inferred, or
that an explicit type be declared for it. This sprouts a lot of additional type annotations
throughout the codebase, including BitcoinTxParser.

Replacement for #74, which I broke with a bad rebase of #71 😒 

See #65.


src/BitcoinHelpers.js
```diff
@@ -344,9 +380,9 @@ const BitcoinHelpers = {
      * @param {number} requiredConfirmations A number of required
      *        confirmations below which this function will return null.
      *
-     * @return {Promise<number>} A promise to the current number of
-     *         confirmations for the given `transaction`, iff that transaction has
-     *         at least `requiredConfirmations` confirmations.
+     * @return {Promise<number?>} A promise to the current number of
+     *         confirmations for the given `transaction`, iff that transaction
```
ironng, line 384: iff -> small typo from previous

src/BitcoinHelpers.js
```diff
@@ -576,8 +618,6 @@ const BitcoinHelpers = {
      * by the `bitcoinAddress`.
      *
      * @param {string} bitcoinAddress Bitcoin address to check.
-     *
-     * @return {Promise<Array>} A promise to the found array of transactions.
```
ironng, line 580: Should this be `@return {Promise<TransactionInBlock[]}` rather than removed?

src/Constants.js
```diff
@@ -63,46 +89,24 @@ class Constants {
     batch.execute()
     const results = await Promise.all(calls)
 
-    const constants = {}
-    results.forEach(constantEntry => {
-      Object.assign(constants, constantEntry)
-    })
+    /** @type {ConstantFields} */
+    const constants = Object.assign({}, ...results)
```
ironng, line 93: 💪

src/Deposit.js
```diff
@@ -133,99 +136,93 @@ export class DepositFactory {
...
+  /**
+   * @private
+   *
+   * Helper to ensure that the contract is defined before returning it. Throws
+   * if the contract is undefined.
+   *
+   * @param {Contract | undefined} contract The contract to check for existence.
+   * @return {Contract} The contract, if defined.
+   */
+  contractOrBust(contract) {
```
ironng, line 148: 😄

src/Deposit.js
```diff
@@ -509,16 +518,19 @@ export default class Deposit {
...
+      .depositToken()
+      .methods.ownerOf(this.address)
       .call()
   }
 
   async inVendingMachine() /* : Promise<boolean>*/ {
     return (
-      (await this.getOwner()) ==
-      this.factory.vendingMachineContract.options.address
+      (await this.getOwner()) == this.factory.vendingMachine().options.address
```
ironng, line 533: I know this is from before, but is this intentionally `==` instead of `===`?

src/Deposit.js
```diff
@@ -1096,22 +1117,37 @@ export default class Deposit {
...
+   * qualification flows.
+   *
+   * @param {{ transactionID: string, outputPosition: number }} bitcoinTransaction
+   *        The transaction id to construct the proof for and the output
+   *        position of interest.
+   * @param {number} confirmations The number of confirmations that the proof
+   *        should show the given transaction has received. Must be >= the
+   *        number of confirmations the transaction has already received.
+   *
+   * @return {Promise<[Buffer,Buffer,Buffer,Buffer,number,Buffer,number,Buffer]>}
```
ironng, line 1145: That's a fun one

src/Redemption.js
```diff
@@ -275,7 +275,7 @@ export default class Redemption {
         // However, constructFundingProof includes it for deposit funding
         // proofs. Here, we filter it out to produce the right set of
         // parameters.
-        ...proofArgs.filter(_ => _ != "output position")
+        ...proofArgs.filter(_ => _ != -1)
```
ironng, line 278: Again, I know this is legacy, but do we want to prefer `!==` instead of `!=` in the spirit of stricter typing?

src/BitcoinHelpers.js
Shadowfiend: I believe this was used intentionally here (to mean [if and only if](https://en.wikipedia.org/wiki/If_and_only_if)).

src/BitcoinHelpers.js
Shadowfiend: I removed it here because TypeScript infers it from the simplicity of the internal invocation. I can add it back if that feels better though.

src/Deposit.js
Shadowfiend: I don't know that it was intentionally that, though I don't think it hurts. Can bump it to a third `=` though :grin:

src/Deposit.js
Shadowfiend: At what point do one's eyes start swirling 🌀🌀

src/Redemption.js
Shadowfiend: Yah can do.

src/BitcoinHelpers.js
Shadowfiend: Fixed upon further reflection.

src/Deposit.js
Shadowfiend: Fix't.

src/Redemption.js
Shadowfiend: Fix't.

src/BitcoinHelpers.js
ironng: Nice. I've never seen that before, thanks!

PR #85, opened 2020-09-02
A few improvements

corollari:
This PR will implement the changes discussed on discord:
- [x] Convert readme to markdown on CI
- [ ] Generate typescript declarations
- [ ] Initial work on improving tests

Currently it's still WIP, I'll add more changes as I work on them

Also, should I split this into multiple PRs? I thought that given that the changes made here are quite small a single PR might work best.


corollari:
Also, currently the typechecks done on the CI are broken, I've tried to replace the jq script with a `--project` flag that points to jsconfig.json but that doesn't work (I don't know why tho)

corollari:
Fixed merge conflicts

corollari:
Note: f2da1abb5bc4606f591d0f877e2572f4db0fd024 is also applied on #91

.github/workflows/node.yml
```diff
@@ -28,9 +28,16 @@ jobs:
...
+    - run: npm run generate:types
     - run: npm test
       env:
         CI: true
+    - name: Install asciidoctor
+      run: sudo apt-get install asciidoctor
+    - name: Install pandoc
+      run: wget https://github.com/jgm/pandoc/releases/download/2.10.1/pandoc-2.10.1-1-amd64.deb && sudo dpkg -i pandoc-2.10.1-1-amd64.deb
+    - name: Convert readme to markdown
+      run: asciidoctor README.adoc -b docbook -o - | pandoc -f docbook -t gfm > README.md
```
Shadowfiend, line 40: I experimented with this a bit the other day---I think the move here is:
```
asciidoctor README.adoc -o - | pandoc -f html -t markdown_strict - -o README.md
```

Going through HTML seems to produce better Markdown output (e.g., it includes the Table of Contents and title) than using Docbook as the intermediary.

README.adoc
```diff
@@ -44,16 +44,9 @@ const tbtc = await TBTC.withConfig({
...
-            "protocol": "ssl"
-        },
-        "testnetWS": {
-            "server": "electrumx-server.test.tbtc.network",
-            "port": 50003,
-            "protocol": "ws"
-        }
+        server: "electrumx-server.test.tbtc.network",
+        port: 8443,
+        protocol: "wss"
```
Shadowfiend, line 49: ❤️

package.json
```diff
@@ -18,14 +18,16 @@
...
+  "types": "types",
   "scripts": {
     "test": "mocha --timeout 10000",
     "lint": "npm run lint:js",
     "lint:fix": "npm run lint:fix:js",
     "lint:js": "npm run lint:js:eslint && npm run lint:js:types",
     "lint:fix:js": "eslint --fix .",
     "lint:js:eslint": "eslint .",
-    "lint:js:types": "!(npx tsc --allowJs --noEmit $(cat jsconfig.json | jq -r '.compilerOptions | to_entries | map([\"--\\(.key)\",.value]) | flatten | join(\" \")') src/**.js bin/**.js examples/**.js | grep \"^\\(src\\|bin\\|test\\|examples\\)/\") # comment any other passed arguments"
+    "lint:js:types": "!(npx tsc --allowJs --noEmit $(cat jsconfig.json | jq -r '.compilerOptions | to_entries | map([\"--\\(.key)\",.value]) | flatten | join(\" \")') src/**.js bin/**.js examples/**.js | grep \"^\\(src\\|bin\\|test\\|examples\\)/\") # comment any other passed arguments",
```
Shadowfiend, line 29: Just to confirm here---ultimately this did end up working?

package.json
```diff
@@ -18,14 +18,16 @@
...
+    "generate:types": "tsc src/*.js --declaration --allowJs --emitDeclarationOnly --outDir types"
```
Shadowfiend, line 30: :sunglasses:

src/lib/ElectrumClient.js
```diff
@@ -1,6 +1,6 @@
 import ElectrumClient from "electrum-client-js"
 import sha256 from "bcrypto/lib/sha256.js"
-import { backoffRetrier } from "./backoff"
+import { backoffRetrier } from "./backoff.js"
```
Shadowfiend, line 3: 😅

test/BitcoinSPVTest.js
```diff
@@ -1,25 +1,19 @@
...
-const ElectrumClient = require("../src/ElectrumClient")
-const config = require("../../../src/config/config.json")
-
-const fs = require("fs")
-const chai = require("chai")
-const assert = chai.assert
+import { BitcoinSPV } from "../src/lib/BitcoinSPV.js"
+import ElectrumClient from "../src/lib/ElectrumClient.js"
+import fs from "fs"
+import chai from "chai"
```
Shadowfiend, line 4: We generally import these at the global level rather than invoking them on `chai`. Is the reason here to satisfy TypeScript? I seem to recall having run into an issue or two with TypeScript, chai, and tests, but remember figuring out how to deal with it… Knowledge that is lost but likely easily found hehe.

src/lib/ElectrumClient.js
```diff
@@ -467,7 +467,7 @@ export default class Client {
  * @param {string} script ScriptPubKey in a hexadecimal format.
  * @return {string} Script hash as a hex string.
  */
-function scriptToHash(script) {
+export function scriptToHash(script) {
```
Shadowfiend, line 470: Wish there was a way to export this for testing without exporting it for everyone else 🤔

Shadowfiend (commented):
Sorry for the long wait here---had a look and left some notes. This is looking awesome overall though, I think when we merge it I'll run a stable release so we have types in the main package.

package.json
corollari: No, I spent several hours trying to get it to work through other methods that don't involve having to parse the .json file directly (would solve the problem on CI while simplifying the command heavily) but I couldn't get it to work, so it's still broken on CI runners.

Probably the easiest way to solve that problem would be patching/upgrading the version of `jq` used on CI, as that is what causes the breakage. If you feel that's a good solution I'll bundle it into this PR.

src/lib/ElectrumClient.js
corollari: I thought of using something like `rewire` but given that this is an internal file which is not meant to be used as part of the library I thought it would be okay to just export it.

What do you think would be the best solution among these?
- Move it into it's own file and test that directly
- Spy on the implementation using `rewire`

test/BitcoinSPVTest.js
corollari: To be honest I don't really remember the exact reason why I went with this. I'll look into doing it in other ways :ok_hand:

src/lib/ElectrumClient.js
Shadowfiend: Sorry, I should have been clearer---I'm fine with it as is, don't think we need to do any additional work right now. I've been thinking about it some more and really, it might even make more sense as a function of `ElectrumClient`, which is where it was in the first place. Either way, I don't think it needs any more work right now, we are past Good Enough™.

package.json
Shadowfiend: Let's actually pull it to a new PR if that's okay---this one's getting big and starting to drift from coherence haha.

.github/workflows/node.yml
corollari: Ah I see, that's probably better than my command since I didn't directly compare the results achieved when using different intermediary formats.

package.json
Shadowfiend: > I spent several hours trying to get it to work through other methods that don't involve having to parse the .json file directly

Side note: I did this when I originally landed on this solution too haha. It's brutal.

corollari: > Let's actually pull it to a new PR if that's okay

On it!

> Side note: I did this when I originally landed on this solution too haha. It's brutal.

Yeah, I spent way too much time trying to get it to work with tsc's `--project`, which seems to be made for this type of use-cases, but while the options were correctly parsed I kept running into problems :confounded:

src/lib/ElectrumClient.js
corollari: Got it! :ok_hand:

PR #86, opened 2020-09-03
Burst of bcoin: Bump bcoin to 2.1.2

Shadowfiend:
The ancient verison of bcoin we were depending on worked just fine, but
a recent bump in tbtc-dapp revealed some issues with newer handling
by webpack and babel of `BigInt` literals, as well as mishandling of bcoin's
browser-specific remappings. Rather than try to escape `create-react-app`'s
straitjacket to try to fix these ourselves, this PR bumps the bcoin
dependency to the latest version, 2.1.2, and makes relevant adjustments
to ensure that transpiles and packages correctly for browser usage in
tbtc-dapp.

This is actively running on the Ropsten release currently deployed to
dapp.test.tbtc.network.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -585,45 +585,45 @@
       },
       "dependencies": {
         "@ledgerhq/devices": {
-          "version": "5.21.0",
-          "resolved": "https://registry.npmjs.org/@ledgerhq/devices/-/devices-5.21.0.tgz",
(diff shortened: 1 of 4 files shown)


ironng (approved):
🤜 🤛 The bump we needed. Nice fix! 🎊

PR #87, opened 2020-09-03
Thomas the Train Engine: If at first you don't succeed, try try again with exponential backoff

Shadowfiend:
Introduce a generalized exponential backoff helper, `backoffRetrier`, use it
for `sendSafelyRetryable`, introduce `EthereumHelpers.callWithRetry` for
use in the minting flow, and add retries to `ElectrumClient`'s
`getTransactionMerkle` function.

This should close keep-network/tbtc-dapp#253 and
close keep-network/tbtc-dapp#263.

### DIFF
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -5,6 +5,8 @@
 /** @typedef { import("web3-utils").AbiItem } AbiItem */
 /** @typedef { import("web3-core").TransactionReceipt } TransactionReceipt */
 
+import { backoffRetrier } from "./lib/backoff"
+
 /**
  * @typedef {object} DeploymentInfo
  * @property {string} address The address a contract is deployed at on a given
@@ -233,29 +235,9 @@ async function sendSafelyRetryable(
   forceSend,
   totalAttempts
 ) {
-  for (let attempt = 1; true; attempt++) {
-    try {
-      console.debug(`sending transaction; attempt number ${attempt}`)
-
-      return await sendSafely(boundContractMethod, sendParams, forceSend)
-    } catch (exception) {
-      if (attempt === totalAttempts) {
-        console.debug(`last attempt ${attempt} failed; throwing exception`)
-        throw exception
-      }
-
-      const backoffMillis = Math.pow(2, attempt) * 1000
-      const jitterMillis = Math.floor(Math.random() * 100)
-      const waitMillis = backoffMillis + jitterMillis
-
-      console.debug(
-        `attempt ${attempt} failed: ${exception}; ` +
-          `retrying after ${waitMillis} milliseconds`
-      )
-
-      await new Promise(resolve => setTimeout(resolve, waitMillis))
-    }
-  }
+  return backoffRetrier(totalAttempts)(async () => {
+    return await sendSafely(boundContractMethod, sendParams, forceSend)
+  })
 }
 
 /**
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -241,6 +241,31 @@ async function sendSafelyRetryable(
 }
 
 /**
+ * Calls the bound contract method (using `.call`, that is as a read-only call)
+ * and retries up to `totalAttempts` number of times, or 3 if unspecified.
+ *
+ * @param {ContractSendMethod} boundContractMethod A bound web3 contract method
+ *        with `estimateGas`, `send`, and `call` variants available.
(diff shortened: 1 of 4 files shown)


Shadowfiend:
@lukasz-zimnoch @ironng interested in some early thoughts, though I haven't had a chance to test here quite yet. I'm spraying out some fixes while they're in my head so I can circle back and do some testing after.

Shadowfiend:
This should be ready for a look.

src/EthereumHelpers.js
```diff
@@ -233,29 +235,34 @@ async function sendSafelyRetryable(
...
+ * @param {Partial<SendOptions>} [sendParams] The parameters to pass to `call`.
+ * @param {number} [totalAttempts=3] Total attempts which should be performed in
+ *        case of an error before rethrowing it to the caller.
+ *
+ * @return {Promise<any>} A promise to the result of sending the bound contract
+ *         method. Fails the promise if gas estimation fails, extracting an
+ *         on-chain error if possible.
+ */
+async function callWithRetry(boundContractMethod, sendParams, totalAttempts) {
+  const retries = typeof totalAttempts === "undefined" ? 3 : totalAttempts
```
ironng, line 258: Could this line be skipped if we just use a default param value for `totalAttempts` like:
```
async function callWithRetry(boundContractMethod, sendParams, totalAttempts = 3) {
```

src/lib/ElectrumClient.js
```diff
@@ -394,11 +395,14 @@ export default class Client {
    *         hexadecimal form.
    */
   async getTransactionMerkle(txHash, blockHeight) {
-    return await this.electrumClient
-      .blockchain_transaction_getMerkle(txHash, blockHeight)
-      .catch(err => {
-        throw new Error(`failed to get transaction merkle: [${err}]`)
-      })
+    return backoffRetrier(3, _ => _.message.includes("not in block at height"))(
```
ironng, line 398: The underscore syntax definitely wins points for being terse but if the error matcher function was written something like
```
err => err.message.includes("my message")
```
it might help with intent. 🤷‍♀️

src/lib/backoff.js
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @template T
+ * @param {number} retries The number of retries to perform before bubbling the
+ *        failure out.
+ * @param {ErrorMatcherFn} [matcher=retryAll] A matcher function that receives
+ *        the error when an exception is thrown, and returns true if
+ * @return {RetrierFn<T>}
+ */
+export function backoffRetrier(retries, matcher) {
+  const errorMatcher = matcher || retryAll // default to always retrying
```
ironng, line 58: Could this also be skipped by using a default value for `matcher` param?

src/lib/backoff.js
```diff
@@ -0,0 +1,88 @@
...
+ *      // message includes "server unavailable".
+ *      backoffRetrier(3, (_) => _.message.includes('server unavailable'))(
+ *        async () => url.get("https://example.com"))
+ *      )
+ *
+ * @template T
+ * @param {number} retries The number of retries to perform before bubbling the
+ *        failure out.
+ * @param {ErrorMatcherFn} [matcher=retryAll] A matcher function that receives
+ *        the error when an exception is thrown, and returns true if
```
ironng, line 54: Is this description incomplete?

ironng (approved):
This looks good! I left a couple questions but nothing merge blocking. 🚋

src/EthereumHelpers.js
Shadowfiend: I need to see when node started supporting that... I think it's too new?

Shadowfiend: (Target being Node 12.)

src/lib/backoff.js
Shadowfiend: 😅

src/lib/ElectrumClient.js
Shadowfiend: Yeah. Basically if it's one statement I always default to _, but agree that here the err would reallllyyyy help. Will make the change.

src/EthereumHelpers.js
ironng: Ah right, good point.

Shadowfiend:
Given the approval, I'm going to merge here to bundle this into a release and make a follow-up PR with fixes.

src/EthereumHelpers.js
Shadowfiend: Confirmed this works :thumbsup:

PR #89, opened 2020-09-04
UTX-Who?: Filter redemption Bitcoin transactions by funding UTXO

Shadowfiend:
During redemption, tbtc.js's `Deposit` object has to identify the Bitcoin
transaction that represents the redemption of this deposit (once it has
been broadcast). The previous strategy was very naive, matching solely
based on the value + receiver script of the transaction output. This meant
that an account redeeming two deposits of the same value to the same
receiver script would lead to confusion in tbtc.js, and potentially submitting
a proof for a transaction that doesn't match the deposit (which fails on the
contract side).

The new approach adds filtering by the outpoint (a transaction id + output
index that identifies the UTXO) used for the transaction input. The outpoint
from the funding UTXO is passed through, and only a transaction that matches
output value, receiver script, and outpoint is matched.

Closes keep-network/tbtc-dapp#348.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -318,18 +318,21 @@ const BitcoinHelpers = {
      *        non-0x-prefixed hex string.
      * @param {number} expectedValue The expected value of the transaction
      *        to fetch.
+     * @param {string} [outpoint] If specified, only finds a transaction that
+     *        spends the specified outpoint.
      *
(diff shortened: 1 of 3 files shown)


ironng (approved):
This is great! Tested a couple runs of a case locally that typically results in the UTXO error, and no error was thrown. 👏👏👏

PR #90, opened 2020-09-09
Thomas cleanup: Few cleanup tasks from Thomas the Train Engine PR

Shadowfiend:
Direct import from #87 .

### DIFF
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -254,9 +254,12 @@ async function sendSafelyRetryable(
  *         method. Fails the promise if gas estimation fails, extracting an
  *         on-chain error if possible.
  */
-async function callWithRetry(boundContractMethod, sendParams, totalAttempts) {
-  const retries = typeof totalAttempts === "undefined" ? 3 : totalAttempts
-  return backoffRetrier(retries)(async () => {
+async function callWithRetry(
+  boundContractMethod,
+  sendParams,
+  totalAttempts = 3
+) {
+  return backoffRetrier(totalAttempts)(async () => {
     // @ts-ignore A newer version of Web3 is needed to include call in TS.
     return await boundContractMethod.call({
       from: "", // FIXME Need systemic handling of default from address.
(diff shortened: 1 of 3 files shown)


ironng (commented):
Looks great so far 💅 

I tried testing this out locally but was hit with a compilation error that prevents the dapp from loading:
```
../tbtc.js/src/lib/backoff.js 66:15
Module parse failed: Cannot use keyword 'await' outside an async function (66:15)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|       try {
|         console.debug(`making attempt number ${attempt}`);
>         return await fn();
|       } catch (error) {
|         if (!errorMatcher(error)) {
```
(I realize this is from the previous PR. Sorry I didn't test it then! 🙈 )

We should be good if return the `await fn()` inside an anonymous async function. Seems like Node doesn't support [top level await until v14.3](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#support-for-top-level-await).

Shadowfiend:
🤔 but this transpiled fine in dev and is running on Ropsten now 🤨🙃

Shadowfiend:
Also this already is inside an anonymous async function (the one returned by `backoffRetrier`. I now remember seeing this issue briefly when I first wrote the code, trying to remember how it got fixed... Feels like it might have been a transpiler issue. I'll poke it some more tomorrow.

ironng:
😕 🙃 🤔  Maybe I need to blow away node modules? I'll retry.

ironng:
I did fresh installs after deleting node modules for both tbtc.js and the dapp and still encountered the compiler error. FWIW I'm running Node v11.15, but like you mentioned, it seems like a transpiler thing.

Shadowfiend:
😭😭😭

Shadowfiend:
Fixed the tranpsilation issue.

ironng (approved):
🚀

PR #91, opened 2020-09-09
12 Nodes A-node-ing: Adjust for Node 12 support

Shadowfiend:
Since Node 12 is the LTS version, let's make sure everything works out of the box with that version and on latest macOS. This requires a few module-loader-related adjustments that may well be eased up in later work, but for now this will do. It also requires moving to a Keep-managed fork of bcoin that points to bcrypto 5.3.0; see bcoin-org/bcoin#993 for more. We'll move back to a proper bcoin version when that gets resolved upstream.

Draft until #90 lands.


Shadowfiend:
Rebased to pull in transpilation fix from #90. This is ready to be undrafted when that PR lands.

ironng (approved):
Looks good! 🕊️🎄🥁

PR #92, opened 2020-09-16
In `purchaseSignerBondsAtAuction` get rightful ETH by calling `withdrawFunds`

daramir:
# Description
- sendSafely action on the `withdrawFunds` method: 
    This way the purchaser will also get the sanctioned ETH as part of `purchaseSignerBondsAtAuction`

## Checklist
- [X] Fork keep-network/tbtc.js
- [X] Clone your fork
- [X] Describe the change you are intending to undertake in the PR description.

Before marking the PR as ready for review, make sure:
- [X] It passes the linter checks.
- [ ] Your changes have sufficient test coverage (e.g regression tests have been added for bug fixes, unit tests for new features)
tests disabled at the moment
- [X] Commits must be signed.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -1310,6 +1310,16 @@ export default class Deposit {
   }
 
   /**
+   * Pays off the deposit balance and closes the liquidation auction
+   * via `Deposit.purchaseSignerBondsAtAuction`, then withdraws the purchased
+   * ETH by calling `DepositUtils.withdrawFunds`.
+   */
+  async takeAuction() {
+    await this.purchaseSignerBondsAtAuction()
+    await this.withdrawFunds()
+  }
+
+  /**
    * Purchases the signer bonds and closes the liquidation auction.
    */
   async purchaseSignerBondsAtAuction() {
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -1331,6 +1331,10 @@ export default class Deposit {
     await EthereumHelpers.sendSafely(
       this.contract.methods.purchaseSignerBondsAtAuction()
     )
+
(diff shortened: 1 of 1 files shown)


Shadowfiend:
So I've been background-thinking about this a bit.

I think the methods named exactly the same as contract methods should probably only invoke those contract methods. Instead of making `purchaseSignerBondsAtAuction` call `withdrawFunds`, what do you think about introducing a separate method, something like `takeAuction` or `closeAuction`, that does both?

We should probably also introduce a wrapper for `withdrawFunds`, similar to the one for `purchaseSignerBondsAtAuction`, that just does a `sendSafely`.

daramir:
That makes sense, @Shadowfiend! Will combine both of your advice and update this PR.

daramir:
@Shadowfiend  PR updated!

Shadowfiend (approved):
Yes yes, let us do it!

Shadowfiend:
Sorry for the wait here <_<

Shadowfiend:
Actually it probably also makes sense to add an `approve` step before `purchaseSignerBondsAtAuction` here, to approve the deposit to spend the sender's BTC.

daramir:
That should be coved by the `purchaseSignerBondsAtAuction` function itself at the moment. I'm happy to create another PR to extract the approve logic from it to `approveTbtcForDeposit`, or just plain `approve`. Is that what you're after @Shadowfiend ?  😁

Shadowfiend:
Oh actually no, I just forgot it was already happening in there 😬

daramir:
Cool, all done then! 🎉

PR #94, opened 2020-09-22
Bumpity Fees: Make a few mainnet adjustments

Shadowfiend:
This includes one revised retry and a doubling of the redemption fee requested
during the redemption flow. Both are based on some less-than-ideal scenarios
seen on mainnet. In particular the redemption fee should be a safer one while support
for user-specified redemption fees is implemented.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -481,7 +481,7 @@ const BitcoinHelpers = {
     estimateFee: async function(tbtcConstantsContract) {
       return new BN(
         tbtcConstantsContract.methods.getMinimumRedemptionFee().call()
-      ).muln(2)
+      ).muln(4)
     },
     /**
      * For the given `transactionID`, constructs an SPV proof that proves it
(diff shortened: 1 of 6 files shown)


src/BitcoinHelpers.js
```diff
@@ -474,11 +475,13 @@ const BitcoinHelpers = {
...
      *
-     * @return {Promise<number>} The estimated fee to execute the provided
+     * @return {Promise<BN>} The estimated fee to execute the provided
      *         transaction.
      */
     estimateFee: async function(tbtcConstantsContract) {
-      return tbtcConstantsContract.methods.getMinimumRedemptionFee().call()
+      return new BN(
+        tbtcConstantsContract.methods.getMinimumRedemptionFee().call()
+      ).muln(2)
```
mhluongo, line 484: I think `4`  would be more appropriate until we have better feerate estimation

mhluongo (changes requested):
LGTM. Let's bump the fee a little higher though

src/BitcoinHelpers.js
Shadowfiend: Done.

PR #96, opened 2020-09-23
Engineering Estimates: Force deposit creation gas limit

Shadowfiend:
This PR deals with an issue where sometimes `estimateGas` returns a
slightly-too-low value for `createDeposit`'s gas needs. `createDeposit`
needs ~1.5M-1.575M gas, but sometimes the esimates bring in a value closer to
1.45M. When these transactions run out of gas, they still eat the full 1.45M
gas limit, which is super lame.

This PR forces the gas limit for deposit creation to 1.6M (subject to user
confirmation of course) to avoid these issues.


mhluongo (approved):
Looks good :ok_hand:

PR #97, opened 2020-09-25
Changing fixed gas estimate for deposit creation to a multipier

dimpar:
This PR is closely related to the changes made in https://github.com/keep-network/tbtc.js/pull/96

Sortition pool updates can mean variation between the required gas at
gas estimation time and at transaction mining time. Rather than try to
set a fixed cost, gas estimation is now followed, but with a multiplier boost
applied on top for deposit creation transactions.

As an addendum, Bitcoin fee estimates, which are still faked as a fixed value,
are now boosted by 4.5x more to account for recent Bitcoin network fee
movement.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -482,7 +482,7 @@ const BitcoinHelpers = {
     estimateFee: async function(tbtcConstantsContract) {
       return new BN(
         await tbtcConstantsContract.methods.getMinimumRedemptionFee().call()
-      ).muln(4)
+      ).muln(18)
     },
     /**
      * For the given `transactionID`, constructs an SPV proof that proves it
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -282,7 +282,9 @@ export class DepositFactory {
 
     const result = await EthereumHelpers.sendSafely(
       this.depositFactory().methods.createDeposit(lotSize.toString()),
-      { value: creationCost, gas: 1700000 }
+      { value: creationCost },
+      false,
+      1.2
     )
 
     const createdEvent = EthereumHelpers.readEventFromTransaction(
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -258,7 +258,7 @@ export class DepositFactory {
 
     const result = await EthereumHelpers.sendSafely(
       this.depositFactory().methods.createDeposit(lotSize.toString()),
-      { value: creationCost, gas: 1600000 }
+      { value: creationCost, gas: 1700000 }
     )
 
     const createdEvent = EthereumHelpers.readEventFromTransaction(
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -205,18 +205,31 @@ function bytesToRaw(bytesString) {
  * method to get the proper underlying error message. Otherwise, sends the
  * signed transaction normally.
  *
+ * Note that if a gas estimate is calculated but is less than an explicitly
+ * specified `sendParams.gas` value, `sendParams.gas` is preferred, and vice
+ * versa. The higher of the two gas values is then multiplied by
+ * `gasMultiplier`.
+ *
  * @param {ContractSendMethod} boundContractMethod A bound web3 contract method
  *        with `estimateGas`, `send`, and `call` variants available.
  * @param {Partial<SendOptions>} [sendParams] The parameters to pass to
  *        `estimateGas` and `send` for transaction processing.
  * @param {boolean} [forceSend] Force the transaction send through even
  *        if gas estimation fails.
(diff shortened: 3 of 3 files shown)


Shadowfiend:
Ok, I did a quick inventory of deposits opened overnight, and some reach a little past 1.7M. I think the move here is going to be to change `sendSafely` to use the explicit gas as a floor rather than an absolute number, and keep the bumped 1.7M floor.

Shadowfiend:
Changed the approach completely. We now trust the gas estimate but give it a 20% boost to account for sortition pool changes between gas estimation and transaction mining.

src/Deposit.js
```diff
@@ -282,7 +282,9 @@ export class DepositFactory {
 
     const result = await EthereumHelpers.sendSafely(
       this.depositFactory().methods.createDeposit(lotSize.toString()),
-      { value: creationCost, gas: 1600000 }
+      { value: creationCost },
+      false,
+      1.2
```
mhluongo, line 287: :+1:

src/EthereumHelpers.js
```diff
@@ -205,26 +205,41 @@ function bytesToRaw(bytesString) {
...
   try {
     // Clone `sendParams` so we aren't exposed to providers that modify `sendParams`.
     const gasEstimate = await boundContractMethod.estimateGas({ ...sendParams })
 
     return boundContractMethod.send({
       from: "", // FIXME Need systemic handling of default from address.
       ...sendParams,
-      gas: Math.max(gasEstimate, (sendParams && sendParams.gas) || 0)
+      gas:
+        Math.max(gasEstimate, (sendParams && sendParams.gas) || 0) *
```
mhluongo, line 241: Feels weird to use JS math for anything these days. For my own edification, how does it deal with floats?

mhluongo (commented):
Concept ACK. Would like to hear about an edge case before merge

src/BitcoinHelpers.js
```diff
@@ -482,7 +482,7 @@ const BitcoinHelpers = {
     estimateFee: async function(tbtcConstantsContract) {
       return new BN(
         await tbtcConstantsContract.methods.getMinimumRedemptionFee().call()
-      ).muln(4)
+      ).muln(18)
```
mhluongo, line 485: :sweat_smile: so I know we're just hacking here, but this is as easy as eg

```
> curl https://api.blockcypher.com/v1/btc/main
{
  "name": "BTC.main",
  "height": 654832,
  "hash": "0000000000000000000c5007a9f2421e75397db74c176fd0ffcf8aa25a12a177",
  "time": "2020-10-31T03:01:57.684544702Z",
  "latest_url": "https://api.blockcypher.com/v1/btc/main/blocks/0000000000000000000c5007a9f2421e75397db74c176fd0ffcf8aa25a12a177",
  "previous_hash": "000000000000000000052e9104d14ef3eeb200fe7d62cac4efe38933ec6181af",
  "previous_url": "https://api.blockcypher.com/v1/btc/main/blocks/000000000000000000052e9104d14ef3eeb200fe7d62cac4efe38933ec6181af",
  "peer_count": 1044,
  "unconfirmed_count": 68673,
  "high_fee_per_kb": 388739,
  "medium_fee_per_kb": 185825,
  "low_fee_per_kb": 80523,
  "last_fork_height": 650491,
  "last_fork_hash": "00000000000000000007a7691b799d895602ce6b1ccd303c6446703a41e6d6de"
}
```

if we want to actually get an estimated feerate in the meantime

src/EthereumHelpers.js
Shadowfiend: Lol. Good catch.

```
ERROR  Error: Error: [number-to-bn] while converting number 2065005.5999999999 to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "2065005.5999999999"
```

🙃

src/BitcoinHelpers.js
mhluongo: Ignoring this and merging for now

Shadowfiend: But then when we update to support multiple Electrum backends, we also need to support multiple centralized APIs for fee rates, or at least warn users that we'll be hitting a single centralized one as a backstop. Neither feels great. I would rather find some good Electrum servers to move between or something of the sort.

TBD in a later PR though.

mhluongo (approved):
:ship:

src/BitcoinHelpers.js
mhluongo: I'll warn you that even with many Electrum servers, we'll likely *still* have to have a list of centralized services we hit up, perhaps randomized, and then user feerate setting... because Bitcoin-land just doesn't have robust infrastructure for this part.

But yeah I hear you on the complexity / wanting to avoid SPoF

PR #98, opened 2020-09-28
Fallaback getEvent to past events querying on subscription error

nkuba:
The change was made after testing tbtc.js `getEvent` function locally.
When subscribing to an event in `once` function the following error was returned:
`The method eth_subscribe does not exist/is not available`. 
As a result `clearInterval` function was executed and the fallback defined in `setInterval` was never executed.


Shadowfiend (approved):
Looks good, let's do it. If I remember correctly, this only happens on non-WebSocket RPCs, fwiw. But since the code is written to work with non-WebSocket anyway, this is perfect.

PR #101, opened 2020-10-15
Sanding Block: Add bin/tbtc.js support for computing digests and broadcasting Bitcoin transactions

Shadowfiend:
It's a bit sloppy right now, but it does the trick. Here is the help doc:

```
    bitcoin digest <transaction-id> <output-index> <output-address>
        Builds a new Bitcoin transaction that can be replaced by fee and outputs the
        digest for that transaction. The transaction id is the id of the transaction
        whose output will be used as an input to this transaction. The output index
        is the index in the referenced transaction that should be used as an input
        to this transaction. The output address should be a bech32 Bitcoin
        address, which will be set up to receive the value at the given transaction
        id and output index, less a Bitcoin transaction fee.

    bitcoin broadcast <transaction-id> <output-index> <output-address> <signer-pubkey> <digest-signature>
        Builds a new Bitcoin transaction that can be replaced by fee and broadcasts
        it to the Bitcoin chain. The transaction id is the id of the transaction
        whose output will be used as an input to this transaction. The output index
        is the index in the referenced transaction that should be used as an input
        to this transaction. The output address should be a bech32 Bitcoin
        address, which will be set up to receive the value at the given transaction
        id and output index, less a Bitcoin transaction fee. The signer pubkey is
        the public key of the signer as an unprefixed 64-byte hex string of
        concatenated x and y coordinates. The digest signature is an unprefixed
        64-byte hex string representation of the signature's r and s values.
```

Already used to push some txes 😎

### DIFF
--- a/bin/commands/bitcoin.js
+++ b/bin/commands/bitcoin.js
@@ -1,0 +1,212 @@
+/** @typedef {import('../tbtc.js').CommandAction} CommandAction */
+/** @typedef {import('../../src/TBTC.js').ElectrumConfig} ElectrumConfig */
+/** @typedef {import('../../src/TBTC.js').Web3} Web3 */
+/** @typedef {import('../../src/TBTC.js').TBTC} TBTCInstance */
+
+import { BitcoinHelpers } from "../../index.js"
+
+import sha256 from "bcrypto/lib/sha256-browser.js"
+const { digest } = sha256
+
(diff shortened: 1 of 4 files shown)


Shadowfiend:
Going to merge this as it's battle-tested if imperfect at this point.

PR #102, opened 2020-10-22
Updated the way outpoint is assembled to find matching bitcoin transaction

nkuba:
Introduced some updates to outpoint assembling when finding matching transactions to match the format of passed expected outpoint.

Sample transaction input data obtained from electrum:
```
txid: 2e679307cf139b5a486511738140e7bb7ea5eec08c0028d6f6a54d656307603b
vout: 1
```

Sample outpoint passed to the function:
```
0x3b600763654da5f6d628008cc0eea57ebbe74081731165485a9b13cf0793672e01000000
```

We convert transaction details to match this format. It is constructed by combining 4 byte `vout` number and `txid`. Bytes are reversed at the end.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -711,8 +711,19 @@ const BitcoinHelpers = {
           const outpointMatches = (
             await electrumClient.getTransaction(tx.tx_hash)
           ).vin.some(({ txid, vout }) => {
-            const inputOutpoint = txid + vout
-            return inputOutpoint == outpoint
+            const voutBuffer = Buffer.alloc(4)
+            voutBuffer.writeUIntBE(vout, 0, 4)
+
+            const actualOutpoint = Buffer.concat([
+              voutBuffer,
+              Buffer.from(txid, "hex")
+            ]).reverse()
+
+            return (
+              actualOutpoint.compare(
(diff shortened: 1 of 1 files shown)


src/BitcoinHelpers.js
```diff
@@ -711,8 +711,19 @@ const BitcoinHelpers = {
           const outpointMatches = (
             await electrumClient.getTransaction(tx.tx_hash)
           ).vin.some(({ txid, vout }) => {
-            const inputOutpoint = txid + vout
```
Shadowfiend, line 714: Man in retrospect this was a real amateur move where I got absolutely eaten by the fact that concatenating a number to a string is valid in JS. Wowie.

Separately, m I a monster for thinking that this is easier to implement by using the buffer solely for big-endianing the vout? In particular, continuing to use string concatenation for the txid + vout combo and continuing to use string comparison for the outpoint equality? Or is that not going to work for other reasons?

nkuba: Additionaly to vout transforming we are also using buffer to reverse bytes in the combined outpoint. I'm not sure which way shall we take to handle these transformations in the most elegant way.

Shadowfiend: Monstrous. Why do people do this to themselves.

Shadowfiend (approved):
Let's do it.

PR #103, opened 2020-10-22
Fixed initial/latest fee obtaining from events

nkuba:
initialFee is first in the array and latest event is last in the array so we had to update how we get them.


Shadowfiend (approved):
I also had to make this adjustment last night. Roll it.

PR #104, opened 2020-10-22
Updated the way to look for past transactions in electrum

nkuba:
Previously we were looking in electrum only for unspent transactions,
which seemed fine for funding but caused problems in redemption. 

There are situations on mainnet where a transaction proof has not been
submitted to the deposit yet, but the output has been spent already.
With this change, we're looking for the whole history of the script
(including mempool) and filter the results of our interest.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -713,6 +713,9 @@ const BitcoinHelpers = {
         })
 
         const matchingOutput = tx.vout.find(({ scriptPubKey, value }) => {
+          // NOTE: We're looking for transactions with value greater or equal
+          // the expected value. This is not looking only for transactions that
+          // value is exactly as expected!
           return scriptPubKey.hex === receiverScript && value >= expectedValue
         })
 
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -705,7 +705,7 @@ const BitcoinHelpers = {
       )
 
       /** @type TransactionInBlock[] */
-      let transactions = []
+      const transactions = []
 
       transactionsWithScript.forEach(tx => {
         if (!tx) {
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -708,12 +708,6 @@ const BitcoinHelpers = {
       const transactions = []
 
       transactionsWithScript.forEach(tx => {
-        if (!tx) {
-          // TODO: Workaround for a case where ElectrumClient.getTransactionsForScript
-          // fails to get transaction details; solve it properly in ElectrumClient.
-          return
-        }
-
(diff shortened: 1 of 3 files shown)


src/BitcoinHelpers.js
```diff
@@ -698,39 +700,82 @@ const BitcoinHelpers = {
       expectedValue,
       outpoint
     ) {
-      const unspentTransactions = await electrumClient.getUnspentToScript(
+      const transactionsWithScript = await electrumClient.getTransactionsForScript(
```
Shadowfiend, line 703: `transactionsForScript` would probably be a better term here to keep with the rest of the verbiage.

src/BitcoinHelpers.js
```diff
@@ -698,39 +700,82 @@ const BitcoinHelpers = {
...
       )
 
-      for (const tx of unspentTransactions.reverse()) {
-        if (tx.value !== expectedValue) {
-          // Skip if the value doesn't match.
-          continue
+      /** @type TransactionInBlock[] */
+      const transactions = []
+
+      transactionsWithScript.forEach(tx => {
```
Shadowfiend, line 710: This is all mutatey and stuff. I'd rather write it in terms of `map` and a series of transformations fwiw, though it's not worth going through the trouble right this moment. It would look something like:

```
transactionsWithScript
    .map(tx => Object.assign({}, tx, { confirmations: tx.confirmations || 0, value: tx.value * ... })
    .sort((txA, txB) => txB.confirmations - txA.confirmations || txB.value - txA.value)
    .filter(({ vin, vout }) => {
      const foundOutpoint = vout.find({ scriptPubKey, value }) =>
        scriptPubKey.hex === receiverScript && value >= expectedValue
      )
      return foundOutpoint && vin.find(({ txid, vout }) => BitcoinHelpers.compareOutpoint(foundOutpoint, BitcoinHelpers.outpointFrom(txid, vout)))
    }
    .map(...) // convert to the final format
```

Leaving this for probably myself haha.

src/lib/ElectrumClient.js
```diff
@@ -451,10 +452,18 @@ export default class Client {
...
+        // This will produce a `undefined` entry in the list that we need to filter
+        // out.
+        .map(txHash => this.getTransaction(txHash).catch(console.error))
     )
 
-    return transactions
+    // Filter out entries for which `getTransaction` failed in the previous step.
+    /** @type {TransactionData[]} */
+    // @ts-ignore We filtered out void entries.
+    const filteredTransactions = transactions.filter(tx => tx)
```
Shadowfiend, line 464: Feels like we could be a little clearer by using `allSettled` and then filtering + logging failed promises, but that's a nitpick.

Shadowfiend (approved):
Solide!

PR #106, opened 2020-10-28
Fallback to querying past events in intervals on subscription error

nkuba:
If there is a problem with a subscription to events we no longer reject promise and clear the interval. In case of subscription problem we want to fallback to the solution where we query past events in intervals

The change was made after testing tbtc.js `getEvent` function locally with rpc endpoint.
When subscribing to an event in `once` function a following error was returned: 
`The method eth_subscribe does not exist/is not available`. 
As a result `clearInterval` function was executed and the fallback defined in `setInterval` was never executed.


Shadowfiend:
I believe this was a duplicate of #98.

PR #107, opened 2020-10-28
Use contracts deployment block to filter past events

nkuba:
We added a `deployedAtBlock` property to a contract object. The `Contract` type now extends web3's Contract Class with a field `deployedAtBlock`. The field is populated with a value reflecting the block number when the contract was deployed.

For the contracts artifacts containing a hash of the transaction when contract was deployed we use this hash to obtain block number from chain. For contracts that are deployed from factory (`Deposit` and `BondedECDSAKeep` contract), we read an event of contract creation and store block number as the deployment block number.

Previously we filtered events starting from block `0` which was overkill for requests sent to endpoints exposed on geth.
With this change we allow `fromBlock` to be passed to get events function so we filter events properly.

### DIFF
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -1,5 +1,5 @@
 /** @typedef { import("web3").default } Web3 */
-/** @typedef { import("web3-eth-contract").Contract } Contract */
+/** @typedef { import("web3-eth-contract").Contract } Web3Contract */
 /** @typedef { import("web3-eth-contract").ContractSendMethod } ContractSendMethod */
 /** @typedef { import("web3-eth-contract").SendOptions } SendOptions */
 /** @typedef { import("web3-utils").AbiItem } AbiItem */
@@ -11,6 +11,8 @@ import { backoffRetrier } from "./lib/backoff.js"
  * @typedef {object} DeploymentInfo
  * @property {string} address The address a contract is deployed at on a given
  *           network.
+ * @property {string} transactionHash The hash of a transaction in which contract
+ *           was deployed on a given network.
  */
 
 /**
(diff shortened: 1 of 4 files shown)


src/EthereumHelpers.js
```diff
@@ -296,19 +317,25 @@ async function callWithRetry(
...
-  contract.options.from = web3.eth.defaultAccount || undefined
+  web3Contract.options.from = web3.eth.defaultAccount || undefined
   // @ts-ignore A newer version of Web3 is needed to include handleRevert.
-  contract.options.handleRevert = true
+  web3Contract.options.handleRevert = true
+
+  /** @type {Contract} */
+  const contract = (web3Contract)
+
+  contract.deployedAtBlock = deployedAtBlock || undefined
```
Shadowfiend, line 338: I would almost prefer if we set this as an `options` value, tbh. That feels like a more appropriate place for this.

src/EthereumHelpers.js
```diff
@@ -288,6 +300,15 @@ async function callWithRetry(
...
 }
 
+/**
+ * @typedef {Object} DeployedContract
+ * @property {number|undefined} deployedAtBlock Number of block when contract was
+ * deployed.
+ *
+ * @typedef {Web3Contract & DeployedContract} Contract web3.eth.Contract enhanced
+ * with property containing block number when contract was deployed.
+ */
```
Shadowfiend, line 310: We should probably define this up top. I would also suggest we name this combined type `DeployedContract` so our imports elsewhere are clearer, and maybe name our adjusted object `ContractDeploymentMixin` or similar.

Shadowfiend (approved):
Haaaate that we've made `getDeployedContract` async, mostly because of how it infects so many other things, but what are you going to do.

PR #108, opened 2020-10-28
Export function to convert deposit state from number to string

nkuba:
The function can be used to convert numeric state values returned by the Deposit contract to a string representation of a state.


Shadowfiend:
This is already exposed as [`tbtc.Deposit.stateById`](https://github.com/keep-network/tbtc.js/blob/master/src/Deposit.js#L94). Is that enough?

PR #109, opened 2020-10-28
Use pWaitFor instead setInterval for get past events in intervals

nkuba:
`setInterval` doesn't fully respect async nature of the action that is
called and doesn't block until the action is fulfilled. When a call to
get events takes long time (longer than 3 seconds) with setInterval we're
starting subsequent calls without waiting for the function to resolve on
the first call. `pWaitFor` will wait for each call to resolve
before starting next action.

### DIFF
--- a/src/EthereumHelpers.js
+++ b/src/EthereumHelpers.js
@@ -6,6 +6,7 @@
 /** @typedef { import("web3-core").TransactionReceipt } TransactionReceipt */
 
 import { backoffRetrier } from "./lib/backoff.js"
+import pWaitFor from "p-wait-for"
 
 /**
  * @typedef {object} DeploymentInfo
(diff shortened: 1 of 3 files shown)


src/EthereumHelpers.js
```diff
@@ -93,27 +94,32 @@ function readEventFromTransaction(
...
     // for events doesn't work correctly we pull past events in a loop until
     // we find our event. This is a temporary solution which should be removed
     // after problem with MetaMask is solved.
     // See: https://github.com/MetaMask/metamask-extension/issues/7270
-    const handle = setInterval(
-      async function() {
+    await pWaitFor(
+      async () => {
         // Query if an event was already emitted after we start watching
-        const event = await getExistingEvent(sourceContract, eventName, filter)
```
Shadowfiend, line 105: Don't love pulling in an extra dependency for this change tbh. Feels like we could write a 15-line helper that would do it that we own and don't need to track dependency updates for.

Shadowfiend (approved):
Let's roll with this for now.

src/EthereumHelpers.js
```diff
@@ -93,27 +94,32 @@ function readEventFromTransaction(
...
+        } catch (error) {
+          console.warn(`failed to get existing event: ${error.message}`)
+        }
 
         if (event) {
-          clearInterval(handle)
           resolve(event)
+          return true
         }
+        return false
```
Shadowfiend, line 117: I guess if `sourceContract.once` resolves the event here, this will run until it also manages to resolve the event, and that second `resolve` will just be a noop.

PR #110, opened 2020-10-29
Expose increase redemption fee method on Redemption class

nkuba:
We exposed function to increase redemption fee on Redemption class, so it can be used in scripts using this library.

Depends on: https://github.com/keep-network/tbtc.js/pull/107

### DIFF
--- a/src/Redemption.js
+++ b/src/Redemption.js
@@ -236,6 +236,49 @@ export default class Redemption {
   }
 
   /**
+   * Increases redemption transaction fee.
+   */
+  async increaseRedemptionFee() {
+    console.log(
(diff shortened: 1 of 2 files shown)


src/Redemption.js
```diff
@@ -235,6 +235,53 @@ export default class Redemption {
     return this.autoSubmittingState
   }
 
+  /**
+   * Increases redemption transaction fee.
```
Shadowfiend, line 239: Worth noting here that the increase is done by a an increment equal to the initial redemption fee, but not a blocker.

Shadowfiend (approved):
👏👏👏👏

The way I wanted to do it in the first place :)

PR #112, opened 2020-10-29
Let redemption auto submit quit if not enough confirmations

nkuba:
Previously the `autoSubmit` function hanged waiting for bitcoin transactions confirmations. With this change we are adding a possibility to quit the flow in case the confirmations number is insufficient.

### DIFF
--- a/src/BitcoinHelpers.js
+++ b/src/BitcoinHelpers.js
@@ -438,31 +438,54 @@ const BitcoinHelpers = {
       onReceivedConfirmation
     ) {
       return BitcoinHelpers.withElectrumClient(async electrumClient => {
-        const checkConfirmations = async function() {
-          return BitcoinHelpers.withElectrumClient(async electrumClient => {
-            const { confirmations } = await electrumClient.getTransaction(
-              transactionID
-            )
+        const handleNewBlock = async function() {
+          const confirmations = await BitcoinHelpers.Transaction.checkConfirmations(
+            transactionID,
+            requiredConfirmations
+          )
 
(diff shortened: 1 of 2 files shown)


src/BitcoinHelpers.js
```diff
@@ -438,31 +438,54 @@ const BitcoinHelpers = {
...
+     *        confirmations.
+     *
+     * @return {Promise<number|null>} A promise to the final number of confirmations
+     *         observed that was at least equal to the required confirmations.
+     *         If the required number of confirmations is not met it will return
+     *         `null`.
+     */
+    checkConfirmations: async function(transactionID, requiredConfirmations) {
+      return BitcoinHelpers.withElectrumClient(async electrumClient => {
+        // TODO: Consider using retry backoff to get transaction.
```
Shadowfiend, line 478: 100%, have this in another branch that improves our retries heh.

Shadowfiend (approved):
I actually question this move. The whole point of `autoSubmit` is that it runs through the whole submission flow end to end.

I think this requirement is best expressed as a new function of some sort, though I acknowledge that'll probably require some refactoring.

I'm guessing the goal here is when one is trying to model “submit a funding proof if possible” as a flow. To me this means the work `broadcastTransactionID` does in `autoSubmit` should be put in its own function, which can be called by an outside caller if they want to get the transaction id. Then they can check confirmations directly, and `autoSubmit` can continue to operate autonomously until completion when called. Thoughts?

Shadowfiend:
Going to hold this guy for now since it's not strictly necessary for tbtc-dapp and the script that is leveraging it is already using this version of the code.

PR #114, opened 2020-12-01
Add `block` parameter to `callWithRetry` function

lukasz-zimnoch:
Added the `block` property to the `callWithRetry` function to allow calling contract methods at a specific block in the past.


Shadowfiend (approved):
Hate that this is necessary, but change looks good!

PR #118, opened 2021-02-16
Error handler for Deposit autoSubmit funding proof error

nkuba:
Previously when an error was returned from `provideBTCFundingProof` transaction in the `autoSubmit` flow an `UnhandledPromiseRejectionWarning` was logged, a user of the library had no chance to handle this error.
```
UnhandledPromiseRejectionWarning: Error: execution reverted: not at current or previous difficulty
    at WebsocketSubprovider._handleSocketMessage (/e2e/node_modules/web3-provider-engine/subproviders/websocket.js:121:18)
    at WebSocket.onMessage (/e2e/node_modules/ws/lib/event-target.js:120:16)
    at WebSocket.emit (events.js:315:20)
    at Receiver.receiverOnMessage (/e2e/node_modules/ws/lib/websocket.js:720:20)
    at Receiver.emit (events.js:315:20)
    at Receiver.dataMessage (/e2e/node_modules/ws/lib/receiver.js:414:14)
    at Receiver.getData (/e2e/node_modules/ws/lib/receiver.js:346:17)
    at Receiver.startLoop (/e2e/node_modules/ws/lib/receiver.js:133:22)
    at Receiver._write (/e2e/node_modules/ws/lib/receiver.js:69:10)
    at writeOrBuffer (_stream_writable.js:352:12)
    at Receiver.Writable.write (_stream_writable.js:303:10)
    at TLSSocket.socketOnData (/e2e/node_modules/ws/lib/websocket.js:795:35)
    at TLSSocket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:302:12)
    at readableAddChunk (_stream_readable.js:278:9)
    at TLSSocket.Readable.push (_stream_readable.js:217:10)
    at TLSWrap.onStreamRead (internal/stream_base_commons.js:186:23)
```

As auto submit happens asynchronously we need a way to bubble up the error. 
In this proposed approach we catch the error and emit it as an event. User is required to define error callback with `onError` function.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -583,6 +583,34 @@ export default class Deposit {
     )
   }
 
+  // /---------------------------- Error Handler -----------------------------
+
+  /**
+   * Emits an event with an error.
+   *
+   * @param {Error} err Error to emit.
+   */
+  notifyError(err) {
+    this.eventEmitter.emit("error", err)
+  }
+
+  /**
+   * A handler that is notified whenever an error occurs.
+   *
+   * @callback OnErrorHandler
+   * @param {Error} err Error
+   */
+
+  /**
+   * Registers a handler for errors.
+   *
+   * @param {OnErrorHandler} onErrorHandler
+   *        A handler that receives an error.
+   */
+  onError(onErrorHandler) {
+    this.eventEmitter.on("error", onErrorHandler)
+  }
+
   // /---------------------------- Event Handlers -----------------------------
 
   /**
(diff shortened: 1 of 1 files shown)


src/Deposit.js
```diff
@@ -1019,8 +1044,8 @@ export default class Deposit {
             {},
             true
           )
-        }
-      )
+        })
+        .catch(this.notifyError)
```
Shadowfiend, line 1048: Why not take an `onError` or similar parameter to `autoSubmit`? It could be optional if someone wanted to handle errors directly on the promise, and we could also emit the event.

In general, a hybrid mechanism where we move to promise workflow + events depending on what the consumer wants to use might be a nice transition for a future direction.

Shadowfiend:
Merging this as-is, but flagging that I'm still interested in the parameter being passed in addition to the notification approach.

src/Deposit.js
nkuba: Great point. Handled in https://github.com/keep-network/tbtc.js/pull/119

PR #119, opened 2021-03-22
Register onError handler with Deposit autoSubmit function

nkuba:
This PR is a follow-up from https://github.com/keep-network/tbtc.js/pull/118/files#r578527898.

We let `onError` handling function be provided with `autoSubmit` function so it's not necessary to call a separate function to register it. The argument is optional.

### DIFF
--- a/src/Deposit.js
+++ b/src/Deposit.js
@@ -601,16 +601,6 @@ export default class Deposit {
    * @param {Error} err Error
    */
 
-  /**
-   * Registers a handler for errors.
-   *
-   * @param {OnErrorHandler} onErrorHandler
-   *        A handler that receives an error.
-   */
-  onError(onErrorHandler) {
-    this.eventEmitter.on("error", onErrorHandler)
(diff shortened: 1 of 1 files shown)


src/Deposit.js
```diff
@@ -1014,16 +1004,23 @@ export default class Deposit {
...
-  autoSubmit() {
+  autoSubmit(onError) {
     // Only enable auto-submitting once.
     if (this.autoSubmittingState) {
       return this.autoSubmittingState
     }
+
+    if (onError) {
+      this.eventEmitter.on("error", onError)
+    }
```
Shadowfiend, line 1022: This isn't what I'd call the expected behavior, fwiw, because this keeps the error handler registered. Passing it to the function implies a single-shot usage. You could use `.one` here, but if there's no error invoked later on then that could lead to an unexpected call with a future error event.

I would just invoke this in the same place we trigger the event below, if it's defined.

PR #121, opened 2021-04-02
Update node workflow and dependencies

michalinacienciala:
We're updating dependencies to `keep-ecdsa` and `tbtc` to use correct
packages for building of `tbtc.js` package. At the moment we use
hardcoded values. In the future we want to make it more flexible and use
latest dependencies automatically.

This PR also changes the workflow for building node dependencies in
`tbtc.js` in order to make it consistent with changes applied in other
GH Actions workflows as part of work on RFC-18 (moving release process
from CircleCI to GitHub Actions).

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,75 +1,99 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-name: Node.js CI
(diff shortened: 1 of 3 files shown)


michalinacienciala:
I removed now the `--dry-run` option from step uploading npm package. I also removed the conditions allowing for testing of the workflow before merge to master. Successfull execution of workflow wth `--dry-run` option can be found here: https://github.com/keep-network/tbtc.js/actions/runs/711531285.

nkuba (approved):

PR #125, opened 2021-05-26
Tweaks to deposit and redemption CLI

nkuba:
This PR introduces a couple of tweaks to deposit creation and redemption with tbtc.js CLI. 

Without the change from c4c7e95 execution of redemption didn't work.
The remaining commits are here to fix some problems with logging.

### DIFF
--- a/bin/commands/deposit.js
+++ b/bin/commands/deposit.js
@@ -520,6 +520,8 @@ async function redeemDeposit(tbtc, deposit, redemptionInfo) {
  * @return {Promise<string>}
  */
 async function runDeposit(tbtc, deposit, mintOnActive) {
+  deposit.onError(console.error)
+
   deposit.autoSubmit()
 
   return new Promise(async (resolve, reject) => {
--- a/bin/commands/deposit.js
+++ b/bin/commands/deposit.js
@@ -502,8 +502,10 @@ async function redeemDeposit(tbtc, deposit, redemptionInfo) {
       }
       redemption.autoSubmit()
 
-      redemption.onWithdrawn(transactionID => {
-        resolve(standardDepositOutput(tbtc, deposit) + "\t" + transactionID)
+      redemption.onWithdrawn(async transactionID => {
+        resolve(
+          (await standardDepositOutput(tbtc, deposit)) + "\t" + transactionID
+        )
       })
     } catch (err) {
       reject(err)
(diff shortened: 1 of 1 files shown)


bin/commands/deposit.js
```diff
@@ -242,7 +242,7 @@ const commandParsers = {
     } else {
       return async tbtc => {
         const deposit = await tbtc.Deposit.withAddress(depositAddress)
-        return redeemDeposit(tbtc, deposit, "")
+        return redeemDeposit(tbtc, deposit, redemptionBitcoinAddress)
```
beaushinkle, line 245: To understand what this was doing, I dug into the source for `redeemDeposit`.

```
      let redemption
      if (typeof redemptionInfo == "string") {
        redemption = await deposit.requestRedemption(redemptionInfo)
      } else {
```

Did this work at all before?

beaushinkle:
Okay - code is simple enough; just the small question about past behavior.

I don't think we should block this PR, but what do you think about adding some backlog work to get some integration testing set up for these commands? Something like https://medium.com/@zorrodg/integration-tests-on-node-js-cli-part-1-why-and-how-fa5b1ba552fe

nkuba:
> I don't think we should block this PR, but what do you think about adding some backlog work to get some integration testing set up for these commands?

💯💯 💯 I'm a huge fan of adding tests for everyting that can be tested! 😆

bin/commands/deposit.js
```diff
@@ -502,8 +502,10 @@ async function redeemDeposit(tbtc, deposit, redemptionInfo) {
       }
       redemption.autoSubmit()
 
-      redemption.onWithdrawn(transactionID => {
-        resolve(standardDepositOutput(tbtc, deposit) + "\t" + transactionID)
+      redemption.onWithdrawn(async transactionID => {
+        resolve(
+          (await standardDepositOutput(tbtc, deposit)) + "\t" + transactionID
```
Shadowfiend, line 507: Random note because I'm here---there's no need to `await` when passing a promise to `resolve`, as resolving with a promise will wait to resolve the outer promise until the inner promise is resolved (i.e., it flattens it out). See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve on resolving with `then`-ables.

nkuba: Thanks for sharing that! Would it resolve correctly also here where we want to add a string to the result of `await standardDepositOutput(tbtc, deposit)`?

PR #128, opened 2021-07-22
Update custom actions paths after moving actions to `ci`

michalinacienciala:
We've moved custom actions used in the Keep/tBTC Continuous Integration
from separate repositories to `ci` repository (under `actions` folder).
We now need to update all the actions' invokations.

Note that temporarily we call the actions on the feature branch - this
is only for testing purposes and will be changed to `@v1` once tested.

Refs:
https://github.com/keep-network/ci/pull/11
https://github.com/keep-network/keep-core/pull/2545
https://github.com/keep-network/keep-ecdsa/pull/868
https://github.com/keep-network/tbtc/pull/816
https://github.com/keep-network/tbtc-dapp/pull/400
https://github.com/keep-network/coverage-pools/pull/167

TODO:
- [x] test deployment flow on the feature branch
- [x] point to `@v1` in lines invoking actions (same change required in other repos)

PR #133, opened 2021-08-04
Fix tests

tomaszslabon:
Depends on https://github.com/keep-network/tbtc.js/pull/128.

This PR introduces a few changes that fix tests:
- updated mocha version to `9.0.3` as new mocha versions support ES6 modules
- uncommented all test files
- changes all tests into ES6 modules (replaced `require`'s with `import`'s), as the implementation files were already ES6 modules and new mocha and node versions support ES6 modules
- renamed `test/AddressTest.js`to `test/BitcoinHelpers.js` as functions tested there were moved to `BitcoinHelpers.js`

Note: the test should work with node 14 and higher


michalinacienciala:
In #128 we've introduced `continue-on-error` for the step running tests in the `Node.js` workflow, to ignore failures that are going to be fixed by the #133 PR.

If #128 will get merged before #133, we'll need to remove the `continue-on-error` in the `Run tests` step of `Node.js` workflow as part of #133 work.
If #133 will get merged before #128, we'll need to remove the `continue-on-error` in the `Run tests` step of `Node.js` workflow as part of #128 work.

michalinacienciala (approved):
Looking good from my perspective.

test/BitcoinHelpersTest.js
```diff
@@ -1,51 +1,45 @@
-/*
-const BitcoinAddress = require("../src/Address")
-const Network = BitcoinAddress.Network
-
-const chai = require("chai")
-const assert = chai.assert
+import BitcoinHelpers from "../src/BitcoinHelpers.js"
+import { assert as _assert } from "chai"
+const assert = _assert
```
Shadowfiend, line 3: Why not just `import { assert } from "chai"`?

test/BitcoinHelpersTest.js
```diff
@@ -1,51 +1,45 @@
...
+  describe("publicKeyToP2WPKHAddress", async () => {
     // Concatenated `x` and `y` coordinates derived from a public key specified in
     // [BIP-173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki):
     // 0279BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798
     const publicKey =
       "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8"
 
     it("calculates testnet address", async () => {
-      const chainType = Network.testnet
+      const chainType = "testnet"
```
Shadowfiend, line 14: This should use [`BitcoinNetwork.TESTNET`](https://github.com/keep-network/tbtc.js/blob/27f24cfb4a4ae7701593040a9dec31924500eaac/src/BitcoinHelpers.js#L23).

test/BitcoinHelpersTest.js
```diff
@@ -1,51 +1,45 @@
...
         publicKey,
         chainType
       )
-
       assert.equal(result, expectedResult)
     })
 
     it("calculates mainnet address", async () => {
-      const chainType = Network.mainnet
+      const chainType = "main"
```
Shadowfiend, line 25: Same note re the network constants.

test/ElectrumClientTest.js
```diff
@@ -1,20 +1,18 @@
-/*
-const ElectrumClient = require("../src/ElectrumClient")
-const fs = require("fs")
-const chai = require("chai")
-const assert = chai.assert
-const config = require("../../../src/config/config.json")
+import Client, { scriptToHash } from "../src/lib/ElectrumClient.js"
```
Shadowfiend, line 1: Could we rename this as `import Client as ElectrumClient, { scrip...`? I think it'll be clearer. Or we could change the exported class name in `ElectrumClient.js`, which would be even better IMO.

src/lib/ElectrumClient.js
```diff
@@ -472,7 +472,7 @@ export default class Client {
  * @param {string} script ScriptPubKey in a hexadecimal format.
  * @return {string} Script hash as a hex string.
  */
-function scriptToHash(script) {
+export function scriptToHash(script) {
```
Shadowfiend, line 475: I wonder if we rename this to `scriptToElectrumHash` for clarity--and maybe even make it a `static` method on the default export class?

Shadowfiend (approved):
This looks awesome! Would love to get some of the notes I left addressed, but I think this is also a material improvement without them, so up to y'all if you have the time.

test/BitcoinHelpersTest.js
tomaszslabon: Yes, it works. Fixed it in https://github.com/keep-network/tbtc.js/pull/133/commits/efe26da71067f1bc7120a7086b6b5f55b2e62ea8.

test/BitcoinHelpersTest.js
tomaszslabon: Fixed in https://github.com/keep-network/tbtc.js/pull/133/commits/bc1d9465eb484f68fd39d0f4416d3c4b9b501cc0.

test/BitcoinHelpersTest.js
tomaszslabon: Fixed in https://github.com/keep-network/tbtc.js/pull/133/commits/bc1d9465eb484f68fd39d0f4416d3c4b9b501cc0.

src/lib/ElectrumClient.js
tomaszslabon: That's a good idea, I made it a static method inside `Client`:
https://github.com/keep-network/tbtc.js/pull/133/commits/6616442a4c858f9defd94459368acfcec81b6526

test/ElectrumClientTest.js
tomaszslabon: I opted for simply importing the class with the name `ElectrumClient` (in https://github.com/keep-network/tbtc.js/pull/133/commits/75fa11240670cca56ca39e566d4b5a472e2115a1). There is a third party library class that is named `ElectrumClient` in `ElectrumClient.js` and adding another class with that name there could cause confusion.

test/BitcoinHelpersTest.js
Shadowfiend: Whoops, but looks like this file didn't change in that commit :grimacing:

tomaszslabon: Yes, I omitted one import. Fixed in https://github.com/keep-network/tbtc.js/pull/133/commits/e94380a3223b8db8feca53ec0c4f9e1d50d69773.

PR #138, opened 2021-08-31
Support EIP-155

nkuba:
We update dependencies to the latest versions to support EIP-155 replay-protected transactions.

`PrivateKeyWalletSubprovider` expects `chainId` to be provided as a constructor argument. If the value is not provided the provider assumes it deals with mainnet (`chainId = 1`).

After EIP-155 the `chainId` is included when calculating a signature, and transactions that are sent to the Ethereum endpoint with non-matching `chainId` are rejected with `invalid sender`.

The previous solution worked fine with Ethereum endpoint that had replay protection disabled, but once the protection is enabled we need to include the correct `chainId` in the transaction signature.


dimpar:
Do want to add `chainId` to /examples dir for `deposit.js` and `redeem.js`?

nkuba:
> Do we want to add a `chainId` to /examples dir for `deposit.js` and `redeem.js`? The rest looks good to me

Nice catch @dimpar! I included it in examples.

dimpar (approved):

PR #146, opened 2022-01-12
Bump-up version of keep-ecdsa and tbtc dependency, add NPM workflow

michalinacienciala:
We recently published first `-dev` suffixed npm package for
`@keep-network/tbtc` (`1.1.2-dev.0`) and bumped-up version of
`@keep-network/keep-ecdsa` package (to `1.9.0-dev.0`)
Now we want to update the dependencies in  the `tbtc.js` to
reflect that change.

We're also adding a NPM workflow similar to what we have in some other
repositories. Workflow will be executed whenever a change in npm 
dependencies or in `bin/`, `src/` or `index.js` occurs on `main`. It will
be also possible to execute workflow on demand.
Workflow will update the dependencies to `keep-network` packages and
publish the npm package to the npm registry under bumped-up version
(using `-dev.X` suffix).
Published package will also be tagged with the `development` tag.

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -1,0 +1,47 @@
+name: NPM
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "bin/**"
+      - "src/**"
+      - "index.js"
+      - "package.json"
+      - "package-lock.json"
+  workflow_dispatch:
+
+jobs:
+  npm-publish:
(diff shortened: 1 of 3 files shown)


michalinacienciala:
NPM workflow tested with `--dry-run` on npm publish step here: https://github.com/keep-network/tbtc.js/runs/4800157644?check_suite_focus=true.

nkuba (approved):

